PR #35, opened 2021-04-27
feat: add /write API to CLI contract

danxmoran:
Pretty straightforward.

I added `format: byte` to the request body to hint to codegen that it should use `[]byte` instead of `string`, which simplified porting the `influx write` command to the new repo.


glinton (approved):
looks great

PR #37, opened 2021-04-27
Update swaggerV1Compat.yml

sanderson:
Add note about timestamps in CSV-encoded query results

### DIFF
--- a/contracts/swaggerV1Compat.yml
+++ b/contracts/swaggerV1Compat.yml
@@ -119,7 +119,7 @@ paths:
           name: Accept
           schema:
             type: string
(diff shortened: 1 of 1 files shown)


glinton (approved):
:tada:

contracts/swaggerV1Compat.yml
```diff
@@ -119,7 +119,7 @@ paths:
           name: Accept
           schema:
             type: string
-            description: Specifies how query results should be encoded in the response.
+            description: Specifies how query results should be encoded in the response. **Note:** When using `application/csv`, query results include epoch timestamps instead of RFC3339 timestamps.
```
sanderson, line 122:
```suggestion
            description: Specifies how query results should be encoded in the response. **Note:** With `application/csv`, query results include epoch timestamps instead of RFC3339 timestamps.
```

PR #40, opened 2021-04-30
fix: remove unsupported PostSetupUser route from OSS API

danxmoran:
Pre-work for influxdata/influxdb#21201. The API's implementation will remain in the server code for at least 1 more patch release, and influxdata/influxdb#21345 is marking the API as deprecated in the swagger.yml copy still tracked in OSS.


glinton (commented):
Also need to remove the `src/common/paths/setup_user.yml` file.

glinton (approved):

PR #41, opened 2021-05-04
chore: update static legend and remove deprecated references to legend

TCL735:
In support of https://github.com/influxdata/ui/issues/806

- Changes allow saving a hidden static legend without nulling
- Remove deprecated "Legend"


src/common/schemas/GaugeViewProperties.yml
```diff
@@ -42,9 +42,5 @@
       type: string
     tickSuffix:
       type: string
-    staticLegend:
-      $ref: "./StaticLegend.yml"
```
glinton, line 46: assuming this removal of static legend was intentional

glinton (commented):
Just need to remove the actual `src/common/schemas/Legend.yml` file (assuming nothing else is referencing it).

glinton (approved):
disregard the previous comment, i see the file is removed :man_facepalming:

src/common/schemas/GaugeViewProperties.yml
TCL735: Yes this is intentional. Gauge will not have a static legend.

PR #43, opened 2021-05-04
feat(cli): add /query API to CLI contract

danxmoran:
I had to do some massaging of the existing ymls to make codegen happy, will comment inline with explanations.

### DIFF
--- a/src/cli/paths/query.yml
+++ b/src/cli/paths/query.yml
@@ -1,0 +1,85 @@
+post:
+  operationId: PostQuery
+  tags:
+    - Query
+  summary: Query InfluxDB
+  parameters:
+    - $ref: "../../common/parameters/TraceSpan.yml"
+    - in: header
+      name: Accept-Encoding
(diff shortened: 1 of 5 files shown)


src/cli/paths/query.yml
```diff
@@ -0,0 +1,81 @@
+post:
```
danxmoran, line 1: I had to make a separate `query.yml` for the CLI codegen to use. Nothing new is added here, but the possible input/output types have been cut down to only the things used by the current CLI (flux input, CSV output).

src/common/schemas/OptionStatement.yml
```diff
@@ -4,6 +4,4 @@
     type:
       $ref: "./NodeType.yml"
     assignment:
-      oneOf:
```
danxmoran, line 7: This inline `oneOf` was tripping up the codegen, so I split it into a separate file.

glinton (approved):

PR #49, opened 2021-05-10
chore: add cloud limits and usage APIs

gavincabbage:


gavincabbage:
@glinton These changes were unintended I think, but I'm not sure why they happened

https://github.com/influxdata/openapi/pull/49/files#diff-77716a54c6780002e2ae7c3a1e073c089adca7e3132d01535f1a17f2315c42c4L38-L818

glinton:
ya, looks like it was slightly "broken" before (the schema type wasn't defined in the schema portion of the swagger because in the `-diff` swagger we don't import common types), so those changes can be ignored.
I think iris may want to verify these changes too, since her team consumes this bit as well, if i understand correctly.

src/cloud/paths/orgs_orgID_limits.yml
```diff
@@ -1,27 +1,5 @@
...
-        application/json:
-          schema:
-            $ref: "../schemas/Limit.yml"
-    default:
-      description: unexpected error
-      content:
-        application/json:
-          schema:
-            $ref: "../../common/schemas/Error.yml"
+  $ref: "./orgs_orgID_limits_get.yml"
```
glinton, line 2: this works, but seems too granular. was there a reason to externalize the `get` definition

gavincabbage: Yeah- the existing limits API was only wired under `api/v2private/` and only the GET operation was appropriate for wiring into a public route under `/api/v2/`.

But that said, this reminds me that the response is _slightly_ different, because of [this] piece that causes the public response to be wrapped in the usual wrapper with `links` etc.

So TLDR i think I can just add a new public route and schema for this rather than reusing the private limits bits here.

gavincabbage: > ya, looks like it was slightly "broken" before (the schema type wasn't defined in the schema portion of the swagger because in the `-diff` swagger we don't import common types), so those changes can be ignored.
> I think iris may want to verify these changes too, since her team consumes this bit as well, if i understand correctly.

I'll request review from Iris when I think this is good to go.

glinton (approved):

PR #54, opened 2021-05-12
feat: notebooksd svc

genehynson:
Adds the notebooksd service swagger definition. Hopefully I did it right 😅

### DIFF
--- a/contracts/ref/oss.yml
+++ b/contracts/ref/oss.yml
@@ -5930,6 +5930,9 @@ paths:
       operationId: GetUsers
       parameters:
       - $ref: '#/components/parameters/TraceSpan'
+      - $ref: '#/components/parameters/Offset'
+      - $ref: '#/components/parameters/Limit'
+      - $ref: '#/components/parameters/After'
       responses:
         "200":
           content:
--- a/src/svc-notebooksd.yml
+++ b/src/svc-notebooksd.yml
@@ -26,6 +26,8 @@ components:
       $ref: './svc/notebooksd/schemas/NotebookArray.yml'
     Notebooks:
       $ref: './svc/notebooksd/schemas/Notebooks.yml'
+    Error:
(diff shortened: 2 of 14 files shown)


contracts/ref/oss.yml
```diff
@@ -5930,6 +5930,9 @@ paths:
       operationId: GetUsers
       parameters:
       - $ref: '#/components/parameters/TraceSpan'
+      - $ref: '#/components/parameters/Offset'
+      - $ref: '#/components/parameters/Limit'
+      - $ref: '#/components/parameters/After'
```
genehynson, line 5935: Not sure why this changed. Has nothing to do with notebooksd

glinton: i believe `make generate` wasn't run on a prior pr.. the `ref` directory is a new addition and testing for it will be added soon

src/svc-notebooksd.yml
```diff
@@ -0,0 +1,41 @@
...
+info:
+  title: notebooksd
+  version: 1.0.0
+servers:
+  - url: /
+paths:
+  /flows/orgs/{orgID}/flows:
+    $ref: './svc/notebooksd/paths/flows_orgs_orgID_flows.yml'
+  /flows/orgs/{orgID}/flows/{id}:
+    $ref: './svc/notebooksd/paths/flows_orgs_orgID_flows_id.yml'
```
glinton, line 11: will these routes eventually be deprecated in favor of `/notebooks` and `/notebooks/{id}`?

glinton (approved):
awesome

src/svc-notebooksd.yml
genehynson: yes! They are temporary to make the transition from `flowd` to `notebooksd` easier. Once we move to `notebooksd`, we'll change the UI to use `/notebooks` and I'll remove these `/flows` routes.

PR #58, opened 2021-05-13
fix: use separate models for org POST and PATCH to match actual impl

danxmoran:
Similar to changes made in #36, the spec for patch-/post-ing buckets was reusing the `Organization` schema even though those APIs don't support specifying / overwriting all of the fields that can be in an org.

### DIFF
--- a/src/common/schemas/PostOrganizationRequest.yml
+++ b/src/common/schemas/PostOrganizationRequest.yml
@@ -1,0 +1,7 @@
+type: object
+properties:
+  name:
+    type: string
+  description:
+    type: string
+  required: [name]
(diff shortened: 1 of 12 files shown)


src/common/schemas/PostOrganizationRequest.yml
```diff
@@ -0,0 +1,7 @@
+type: object
+properties:
+  name:
+    type: string
+  description:
+    type: string
+  required: [name]
```
glinton, line 7: surprised there's not a way to require oneof for the patch request.

glinton (approved):

PR #60, opened 2021-05-14
feat(unity): Add additional routes for Usage

gspletzer:
Upon investigating the current set-up for the usage page, it was determined that the `GET usage/{vector_name}` should not include `billing_stats` and `rate_limits` as those queries involve different sections of the page, and do not follow the same schema as `usage_vector`. So, `billing_stats` and `rate_limits` were removed as options for `GET usage/{vector_name}`, and instead two new routes were created to accommodate them: `GET usage/billing_stats` and `GET usage/rate_limits`. Additionally the name of the query param for `GET usage/{vector_name}` and `GET usage/rate_limits` were updated from `start` to `range` for better readability and consistency with Quartz.


contracts/priv/unity.yml
```diff
@@ -144,16 +150,14 @@ paths:
...
-              - billing
               - storage_gb
               - write_mb
               - reads_gb
               - query_count
           description: The name of the UsageVector to query.
           example: reads_gb
         - in: query
-          name: start
+          name: range
```
Asalem1, line 160: nit: Isn’t the range always relative to now? So wouldn’t we just need to provide a start?

Asalem1 (approved):
LGTM

contracts/priv/unity.yml
gspletzer: @Asalem1 It was changed to range in Quartz because start didn't seem as clear for the allowed values of 24h, 7d, and 30d. I suppose you could argue it either way. If it is going to create more work for you, I think I can just change the query param name back to start in Quartz and leave the variable name that acquires that value as range. Let me know what you prefer.

gspletzer:
alchemy please merge

PR #64, opened 2021-05-17
refactor: remove CLI contract and related YMLs

danxmoran:
The CLI contract & custom modifications are being moved into the CLIrepo for better separation of concerns. Common/OSS/Cloud YML components are still being consumed from the new location via the codegen process.

See influxdata/influx-cli#95 for the other side of things.


wbaker85 (approved):
This looks fine to me - taking the CLI-client-specific customization out of the central repo shouldn't impact any other consumers. @glinton might want to take a look as well for awareness.

glinton (approved):

PR #66, opened 2021-05-17
chore: tidy annotations response codes. generalize error responses.

### LINKED ISSUE #65: fix: annotations won't return 405 (method not allowed) on defined methods
Route/method combinations defined in swagger today do not define all the possible error conditions so the 405 response for a good defined method/route combo should be removed, as that response would never trigger for a matching request. Also, the api can 404, but that is not defined.

glinton:
Fixes #65 and removes responses that are only unique in their swagger descriptions as the status codes describe the response well enough.


src/unity/paths/usage_vectors.yml
```diff
@@ -11,10 +11,13 @@ get:
...
     '404':
       description: Pricing version not 4 or User/Org not found
-      $ref: '../../common/responses/ServerError.yml'
+      $ref: '../../common/responses/ServerError.yml
+'
     default:
       description: Unexpected error
-      $ref: '../../common/responses/ServerError.yml'
+      $ref: '../../common/responses/ServerError.yml
+'
```
StacieClark, line 23: Ticks not on the same line....

PR #70, opened 2021-05-19
feat: specify date format for DateTimeLiteral

bednar:
The `DateTimeLiteral` should be specified  with `format` property as a `date-time`. It is useful for generated client because the property will be defined as a `DateTime` not `String`.

image
- https://github.com/influxdata/flux/blob/eb2e578d5af6fc5a6f4a3f8a0ec0a640506f7b1d/ast/ast.go#L1738

image
- https://swagger.io/docs/specification/data-models/data-types/#format

### DIFF
--- a/contracts/cloud.yml
+++ b/contracts/cloud.yml
@@ -7096,6 +7096,7 @@ components:
           $ref: '#/components/schemas/NodeType'
         value:
           type: string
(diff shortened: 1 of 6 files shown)


glinton (approved):
seems reasonable to me

PR #72, opened 2021-05-19
Add a tag to the PostDelete operation.

danxmoran:
This helps codegen systems name the generated API intelligently. Without the tag it ends up in a "default" bucket.


glinton (approved):
lgtm. posting to a `/delete` route would be nice for us to change in the next version of the api if possible

PR #76, opened 2021-05-20
Add missing parameters to users API

danxmoran:
Both OSS and Cloud implement filtering on these parameters.


glinton (approved):
do you know if these are in cloud as well?

danxmoran:
Yes, I see them handled in `tenant/http_server_user.go`

PR #78, opened 2021-05-20
chore: fix close quote mishap (bring back to same line)

glinton:


StacieClark (approved):
Does it bother you that some of the files use single quotes and some use double?

glinton:
>Does it bother you that some of the files use single quotes and some use double?

slightly, but i try not to be too ocd about conglomerate code, otherwise i likely wouldn't sleep well :slightly_smiling_face:. plus there are more pressing concerns than formatting

PR #79, opened 2021-05-20
fix(tasks): add content-type to task retry requests

hoorayimhelping:
Closes influxdata/ui#1475

### DIFF
--- a/src/common/paths/tasks_taskID_runs_runID_retry.yml
+++ b/src/common/paths/tasks_taskID_runs_runID_retry.yml
@@ -3,6 +3,11 @@ post:
   tags:
     - Tasks
   summary: Retry a task run
+  requestBody:
+    content:
+      application/json:
+        schema:
+          $ref: "../schemas/Run.yml"
   parameters:
     - $ref: "../parameters/TraceSpan.yml"
     - in: path
--- a/src/common/paths/tasks_taskID_runs_runID_retry.yml
+++ b/src/common/paths/tasks_taskID_runs_runID_retry.yml
@@ -5,9 +5,9 @@ post:
   summary: Retry a task run
   requestBody:
     content:
-      application/json:
+      application/json; charset=utf-8:
         schema:
-          $ref: "../schemas/Run.yml"
+          type: object
   parameters:
     - $ref: "../parameters/TraceSpan.yml"
     - in: path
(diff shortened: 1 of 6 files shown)


src/common/paths/tasks_taskID_runs_runID_retry.yml
```diff
@@ -3,6 +3,11 @@ post:
   tags:
     - Tasks
   summary: Retry a task run
+  requestBody:
+    content:
+      application/json:
+        schema:
+          $ref: "../schemas/Run.yml"
```
glinton, line 10:
```suggestion
        type: object
```

Unless `Run` is an actual valid body, I'd just specify `type: object` so the headers get set but no default body gets created. Using `type: object` results in the following typescript diff generated by oats (:

```sh
export GEN_TS=true
export DIFF_TS=true
make test-oats
```
```diff
--- orig-ts/common.ts
+++ ts/common.ts
@@ -7365,6 +7365,8 @@
   taskID: string
   runID: string
 
+  data?: any
+
   headers?: {
     'Zap-Trace-Span'?: string
   }
@@ -7393,7 +7395,10 @@
   request(
     'POST',
     `/api/v2/tasks/${params.taskID}/runs/${params.runID}/retry`,
-    params,
+    {
+      ...params,
+      headers: {...params.headers, 'Content-Type': 'application/json'},
+    },
     options
   ) as Promise<PostTasksRunsRetryResult>
 ```

File: src/common/paths/tasks_taskID_runs_runID_retry.yml [hunk 2]
```diff
@@ -3,6 +3,11 @@ post:
   tags:
     - Tasks
   summary: Retry a task run
+  requestBody:
+    content:
+      application/json:
```
glinton, line 8: I'd also consider setting the charset (seems some content type checks are more pedantic than others, annotations i believe currently is)
```suggestion
      application/json; charset=utf-8:
```

```diff
--- orig-ts/common.ts
+++ ts/common.ts
@@ -7365,6 +7365,8 @@
   taskID: string
   runID: string
 
+  data?: any
+
   headers?: {
     'Zap-Trace-Span'?: string
   }
@@ -7393,7 +7395,13 @@
   request(
     'POST',
     `/api/v2/tasks/${params.taskID}/runs/${params.runID}/retry`,
-    params,
+    {
+      ...params,
+      headers: {
+        ...params.headers,
+        'Content-Type': 'application/json; charset=utf-8',
+      },
+    },
     options
   ) as Promise<PostTasksRunsRetryResult>
```

glinton (approved):

PR #81, opened 2021-05-21
fix: change name of NotificationEndpointDiscrimator to NotificationEndpointDiscriminator

bednar:
Fixed typo in name.


glinton (approved):
This looks fine to me, but it will likely require a change to ui as the generated code uses the names from the schema, so someone from that team may want to at least get eyes on this before consuming

Asalem1 (approved):
<3

PR #82, opened 2021-05-21
chore: updated spacing and enum format to make generated types and stuff check out

Asalem1:
Previously generate types were setting `required` as an optional parameter rather than having the nested parameters be required.

In addition, the `enums` in the paths weren't listing an enum, but were rather simply saying that the parameter was a string


ischolten (approved):
👏  Nice cleanup!

PR #86, opened 2021-05-24
chore: made enums their own thing

Asalem1:
An attempt to make the enums their own types since oats is having trouble setting the enum parameters as actual enums. Creating a universal ref for those so that it makes easier to understand the generated code

### DIFF
--- a/src/unity/schemas/Me.yml
+++ b/src/unity/schemas/Me.yml
@@ -6,12 +6,7 @@ properties:
     type: string
     description: the email associated with the user
   accountType:
-    type: string
-    description: type of the account
-    enum:
-      - free
-      - cancelled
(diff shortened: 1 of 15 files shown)


genehynson (approved):

src/unity/schemas/Me.yml
```diff
@@ -6,12 +6,7 @@ properties:
     type: string
     description: the email associated with the user
   accountType:
```
glinton, line 8: maybe:

```suggestion
  accountType:
    type: object
      $ref: './AccountType.yml'
```

glinton:
link https://github.com/influxdata/oats/issues/24 which was noticed after this merged

PR #90, opened 2021-05-25
chore: remove useless RequestBody for `Retry a task run`

bednar:
The request body for `Retry a task run` operation can be omitted. The server implementation only uses query parameters:

https://github.com/influxdata/influxdb/blob/00420fb54c79b9c3ce33b02fe5717a4365a32fe7/http/task_service.go#L1381.

And also the [OpenAPI](https://github.com/OpenAPITools/openapi-generator) generator generate useless model object with name `UNKNOWN_BASE_TYPE`.

### DIFF
--- a/contracts/cloud.yml
+++ b/contracts/cloud.yml
@@ -4457,11 +4457,6 @@ paths:
       tags:
         - Tasks
       summary: Retry a task run
-      requestBody:
-        content:
-          application/json; charset=utf-8:
-            schema:
-              type: object
       parameters:
         - $ref: '#/components/parameters/TraceSpan'
         - in: path
--- a/contracts/common.yml
+++ b/contracts/common.yml
@@ -4457,11 +4457,6 @@ paths:
       tags:
         - Tasks
       summary: Retry a task run
-      requestBody:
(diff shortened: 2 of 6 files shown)


Asalem1:
This isn't useless unfortunately. It was recently added in due to CSRF protections that the API had added. Unfortunately, the tool the UI uses to generate functions `oats` isn't suited to modify content type on the UI so we would need to explicitly set that here

Asalem1 (changes requested):
Here's the original PR with a link to the original issue:

https://github.com/influxdata/openapi/pull/79/files

glinton:
Asalem1 is correct, `RequestBody` was added so oats would always set the `content-type` on the request. Adding it as a parameter was attempted, but oats didn't populate the default value so the `RequestBody` version was adopted.

```yaml
    - in: header
      name: Content-Type
      schema:
        type: string
        default: 'application/json; charset=utf-8'
```

I'm wondering if setting the `Content-Type` parameter as `required` would help in any way.

bednar:
@glinton, @Asalem1 thanks for explanation

Issue #92, opened 2021-05-25
Missing `RunID` on `LogEvent`


lesam:
The `LogEvent` log message for flux task runs contains a `RunID` field in practice, but the swagger spec does not capture this.

PR #93, opened 2021-05-25
fix: add runID to LogEvent

lesam:
Closes https://github.com/influxdata/openapi/issues/92


src/common/schemas/Run.yml
```diff
@@ -23,14 +23,7 @@
...
       items:
-        type: object
-        properties:
-          runID:
-            type: string
-          time:
-            type: string
-          message:
-            type: string
+        $ref: '#/components/schemas/LogEvent'
```
danxmoran, line 26: This looks like the change that's failing CI, another `make generate` should fix it.

lesam: Figured it out - should pass now

danxmoran (approved):

glinton (approved):

PR #100, opened 2021-05-27
feat(unity): update invoice to have id instead of filesId

ischolten:
To fix a bug involving invoices, the api needs to use the id of the invoice rather than the id of the invoice pdf file.


Asalem1 (approved):
💯

PR #108, opened 2021-06-02
chore: add `notebooks` permission resource

bednar:
The InfluxDB OSS has also `notebooks` permission resource type:

https://github.com/influxdata/influxdb/blob/40897b9a98fe48834bf2a6a9469aae89f85a0d33/authz.go#L139

### DIFF
--- a/contracts/oss-diff.yml
+++ b/contracts/oss-diff.yml
@@ -732,6 +732,7 @@ components:
             - notificationEndpoints
             - checks
             - dbrp
+            - notebooks
         id:
           type: string
           nullable: true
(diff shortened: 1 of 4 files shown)


glinton (approved):
good catch

Asalem1:
I don't think that Notebooks is toggled on in OSS

wbaker85:
As part of notebooks for OSS, it was identified that this would have to be done - see https://github.com/influxdata/influxdb/issues/21491 - but I just haven't gotten around to doing it yet. Notebooks for OSS isn't quite toggled on yet but it will be on our next release so I don't _think_ this will cause any issues right now. What I'll do anyway is toggle notebooks on in OSS before I update the UI that the`influxdb` master branch is pointing at - notebook on OSS is ready to go, so might as well at this point.

PR #109, opened 2021-06-02
Add OSS restore APIs and adjust backup APIs to match

### LINKED ISSUE #97: Design new restore API for OSS 2.1.0
OSS v2.1.0 is adding sqlite as another metadata back-end, so its restore API(s) need to be updated/extended to include that data. Add the path and schema definitions for the new API under `src/oss`.

### LINKED ISSUE #99: Extend OSS 2.1 backup API to include DatabaseInfo needed for `influx restore`
The `POST /api/v2/restore/buckets/:bucketID` API currently expects a serialized `meta.DatabaseInfo` as its request body. We'll likely change that format to a JSON equivalent in #97, but the core information will still be required.

The `GET /api/v2/backup/metadata` API needs updated to return enough information for users to extract the database info without needing to open up the downloaded boltDB.

danxmoran:
Closes #97
Closes #99

The new schemas from `/backup/metadata` are based off of `meta.DatabaseInfo` in `influxdb`


wbaker85 (approved):
👍

PR #112, opened 2021-06-07
chore: summaries should be plain English and have consistent verbs

jstirnaman:
* add summaries where missing
* cleanup summary grammar. 
* cleanup description grammar.

closes https://github.com/influxdata/docs-v2/issues/2666

### DIFF
--- a/contracts/cloud-diff.yml
+++ b/contracts/cloud-diff.yml
@@ -135,13 +135,13 @@ paths:
       operationId: GetSetup
       tags:
         - Setup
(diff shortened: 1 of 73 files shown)


glinton (approved):
looks great, only thing i could suggest would be to unify punctuation (some descriptions end in a period while others do not), but that's not critical

jstirnaman:
> looks great, only thing i could suggest would be to unify punctuation (some descriptions end in a period while others do not), but that's not critical

Thanks, @glinton. Sure. WIP.

PR #114, opened 2021-06-08
fix: revert "chore: add notebooks permission resource (#108)"

glinton:
This reverts commit dab6384fb7108f8cee396cc9d187870c3b0381ea because notebooks/flows is intended to be accessible only via a session (this definition is used by the ui for token generation)


Asalem1 (approved):

glinton:
closing because ui is now using cloud swagger for cloud specific interactions

PR #120, opened 2021-06-09
chore: remove typo error code from annotations api

### LINKED ISSUE #111: chore: ensure annotations' response codes are correct
It was pointed out that annotations' swagger is defining a 404 for posing annotations as well as listing all annotations which seems like a typo and should be verified/removed.

glinton:
closes #111

The 404 when listing may or may not be desired, but i won't remove it unless someone feels we should just return an empty slice. currently a 404 will be returned when the filter parameters result in 0 annotations/streams being found (for `/annotations` and `/streams`

### DIFF
--- a/contracts/priv/annotationd.yml
+++ b/contracts/priv/annotationd.yml
@@ -50,8 +50,6 @@ paths:
           $ref: '#/components/responses/ServerError'
         '401':
           $ref: '#/components/responses/ServerError'
-        '404':
-          $ref: '#/components/responses/ServerError'
         '500':
           $ref: '#/components/responses/ServerError'
     delete:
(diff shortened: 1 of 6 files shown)


saraghds:
I tried hitting the listing endpoint and got 200 instead of 404 when result is empty:

```
AssertionError: 
        failure: https://twodotoh.a.influxcloud.dev.local/api/v2private/annotations
        query: {"startTime":"1990-02-01T07:00:00.000Z","endTime":"1990-02-01T07:00:00.000Z"}
        response body: []

: expected 200 to equal 404
      + expected - actual

      -200
      +404
```
Is this behavior different than what is expected?

saraghds:
> I tried hitting the listing endpoint and got 200 instead of 404 when result is empty:
> 
> ```
> AssertionError: 
>         failure: https://twodotoh.a.influxcloud.dev.local/api/v2private/annotations
>         query: {"startTime":"1990-02-01T07:00:00.000Z","endTime":"1990-02-01T07:00:00.000Z"}
>         response body: []
> 
> : expected 200 to equal 404
>       + expected - actual
> 
>       -200
>       +404
> ```
> 
> Is this behavior different than what is expected?

saraghds (approved):

PR #122, opened 2021-06-10
Split out enums for line-protocol error codes.

danxmoran:
This helps some codegen systems generate true enums.

I thought about reusing the existing `ErrorCode` enum, but the possible values weren't a 1:1 match.


glinton (approved):
:tada:

PR #123, opened 2021-06-10
feat(unity): add/update operator routes and schemas

ischolten:
- updates the existing operator route definitions
- removes unused / dupe routes
- updates the OrgLimits schema to fix a structure problem (required is not supposed to be a property)
- the nested responses from the operator ui is a bit yuck, but for the purpose of exposing these routes for the unity epic, we are not changing their functionality or adding new routes, just documenting them as a part of this swagger.

@Asalem1 there are a lot of changes here that will change the code after the ui regenerates so head up. Ill try to mark the big changes as comments in the diff.


contracts/priv/quartz-oem.yml
```diff
@@ -368,15 +368,15 @@ components:
...
           $ref: '#/components/schemas/NotificationEndpointLimits'
-        required:
-          - orgID
-          - rate
-          - bucket
-          - task
-          - dashboard
-          - check
-          - notificationRule
-          - notificationEndpoint
```
ischolten, line 379: quartz-oem changed because OrgLimits was updated to fix a problem where required was nested under properties instead of a sibling of properties

src/unity.yml
```diff
@@ -49,16 +45,22 @@ paths:
     $ref: './unity/paths/orgs_orgId_users_userId.yml'
   '/orgs/{orgId}/limits':
     $ref: './unity/paths/orgs_orgId_limits.yml'
-  '/orgs/{orgId}':
-    $ref: './unity/paths/orgs_orgId.yml'
```
ischolten, line 53: this route is unused and kind of a dupe of the `GET /operator/orgs/{orgId}` route

src/unity.yml
```diff
@@ -112,12 +116,12 @@ components:
       $ref: './unity/schemas/Invoice.yml'
     Invoices:
       $ref: './unity/schemas/Invoices.yml'
-    User:
-      $ref: './unity/schemas/User.yml'
+    OrgUser:
```
ischolten, line 119: the response from the users endpoints on the users page returns what we refer to as OrgUsers vs the operator user schemas which reflect quartz users. There are some differences with those schemas so they were separated.

src/unity/paths/operator_account_accountId.yml
```diff
@@ -1,48 +0,0 @@
-# Oats is making this incredibly annoying with naming conflicts
-get:
```
ischolten, line 2: this just got renamed to operatr_accounts_accountId.yml

src/unity/paths/operator_accounts_accountId.yml
```diff
@@ -0,0 +1,87 @@
...
+      description: The ID of the account to get.
+  responses:
+    '200':
+      description: the account
+      content:
+        application/json:
+          schema:
+              allOf:
+                - $ref: '../schemas/Account.yml'
+                - properties:
```
ischolten, line 23: the operator get account returns an account with layers of nested values nested values
account
  - billing contact
  - users
  - marketplace subscription
  - organizations
    - each org has its account nested

src/unity/paths/operator_orgs_orgId.yml
```diff
@@ -0,0 +1,36 @@
...
+      description: the idpe id of the org
+  responses:
+    '200':
+      description: Organization
+      content:
+        application/json:
+          schema:
+            allOf:
+              - $ref: '../schemas/Organization.yml'
+              - properties:
```
ischolten, line 23: get org returns an org with its account nested

src/unity/paths/operator_orgs_orgId_limits.yml
```diff
@@ -0,0 +1,70 @@
...
+        type: string
+      required: true
+      description: The IDPE id of the org
+  responses:
+    '200':
+      description: Organization Limits
+      content:
+        application/json:
+          schema:
+            $ref: '../../quartz/schemas/OrgLimits.yml'
```
ischolten, line 21: i didnt want to remake the org limits but i dont think we can move it into common yet because there is are slightly different org limits for idpe. so just referencing them here for now.

src/unity/schemas/Account.yml
```diff
@@ -1,20 +1,9 @@
 properties:
   id:
-    type: string
-    description: account id
-  quartzId:
```
ischolten, line 5: when a resource comes from quartz, the id is the quartz id and the other id is the idpe id

src/unity/schemas/Account.yml
```diff
@@ -1,20 +1,9 @@
...
     type: number
     description: account id in quartz
-  onboardingState:
```
ischolten, line 8: onboarding state is related to a user not account (for now)

src/unity/schemas/Account.yml
```diff
@@ -1,20 +1,9 @@
...
-    type: string
-    description: stage of onboarding the account is in
-  marketplace:
```
ischolten, line 11: making a distinction between marketplace (info about a marketplace) and a marketplace subscription (info about a user / accounts subscription to a marketlpace)

src/unity/schemas/Account.yml
```diff
@@ -24,11 +13,4 @@ properties:
...
     description: remaining balance on the account, nil if none
-  users:
-    type: array
-    items:
-      $ref: './User.yml'
-  billingContact:
-    $ref: './BillingContact.yml'
-    description: billing contact for the account
-required: [id, balance, billingContact, marketplace, users, type]
+required: [id, balance, type]
```
ischolten, line 16: different places where accounts are used have different required fields and nested resource. those are defined in the location of use

src/unity/schemas/Accounts.yml
```diff
@@ -1,2 +1,15 @@
 items:
-  $ref: './Account.yml'
+  allOf:
+    - $ref: './Account.yml'
+    - properties:
```
ischolten, line 4: operator get accoutns api returns a list of accounts with some nested resource

account
  - billing contact
  - users
  - marketplace subscription

src/unity/schemas/Organizations.yml
```diff
@@ -1,3 +1,13 @@
 type: array
 items:
-  $ref: './Organization.yml'
+  allOf:
+  - $ref: './Organization.yml'
+  - properties:
```
ischolten, line 5: operator get organizations returns a list of orgs with nested resources

- org
  - account
    - billing contact

src/unity/paths/me.yml
```diff
@@ -1,7 +1,7 @@
 get:
   operationId: GetMe
   tags:
-    - Account
+    - Accounts
```
randycoulman, line 4: Do we want this to be plural given that `Region` and `Organization` below are singular? And for consistency, the two instances above as well?

src/unity/paths/operator_accounts_accountId.yml
```diff
@@ -0,0 +1,87 @@
...
+        type: string
+      required: true
+      description: The ID of the account to get.
+  responses:
+    '200':
+      description: the account
+      content:
+        application/json:
+          schema:
+              allOf:
```
randycoulman, line 21: This is a great solution! Does oats support `allOf`? I know the Elixir OpenAPI generators I looked at a while back don't, which is why I ask.

src/unity/paths/operator_accounts_accountId_users_userId.yml
```diff
@@ -1,9 +1,9 @@
 delete:
-  operationId: RemoveUserFromAccount
+  operationId: removeUserFromAccount
```
randycoulman, line 2: From what I've seen, it looks like operation IDs are typically UppercaseFirst, but I see that a number of them are changed to lowercaseFirst here. Which one should it be?

src/unity/schemas/Organization.yml
```diff
@@ -1,9 +1,9 @@
 type: object
 properties:
-  id:
+  idpeId:
```
randycoulman, line 3: We currently rename this to `idpeID`, so we'll have to remember to remove that rename (and make sure it doesn't break anything, like the public API).  But if the OEM API exposes it as `idpeID`, we'll need to rename it here instead.

Update: just checked, and OEM orgs are quite different than this, so I think we'll be OK as long as we remember to update the code to match this.

randycoulman (approved):
LGTM! A few notes/questions below.

src/unity/paths/operator_accounts_accountId.yml
Asalem1: I believe it does but I'm not certain. It looks like it's used on other openAPI cloud routes so 🤞🏽

src/unity/paths/operator_accounts_accountId.yml
Asalem1: I might be speaking too soon since I haven't gone through all the changes, but I think we currently have a `RelatedAccount` prop that's a limited version of the org's account so that we don't have a circular dependency of `orgs <-> account`. Is that still the case?

**UPDATE:** Just saw that the account was updated to remove the organization <3

src/unity/paths/operator_orgs_orgId.yml
Asalem1: I think this is related to the concern above where an account has an org property and this will be endless

src/unity/paths/operator_orgs_orgId_limits.yml
Asalem1: Out of curiosity, how is the API different for getting the org limits? It looks like our current UI implementation makes a request for an org's limits based on the org_id. If there's no difference, is IDPE or Quartz the source of truth since we should probably reconcile those

src/unity/schemas/Account.yml
```diff
@@ -1,20 +1,9 @@
...
-  onboardingState:
-    type: string
-    description: stage of onboarding the account is in
-  marketplace:
-    $ref: './Marketplace.yml'
-    description: which marketplace, nil if none
   type:
     $ref: './AccountType.yml'
-  organizations:
-    $ref: './Organizations.yml'
```
Asalem1, line 17: Aw yup, there we go. Cool beans

Asalem1 (approved):

src/unity/paths/operator_orgs_orgId_limits.yml
randycoulman: Quartz gets and returns the org limits pretty much directly from IDPE, but I think we leave out a couple of fields we don't care about.

For the operator UI, we definitely need to request these through Quartz, though, because the operator doesn't have permissions on their own to access org limits for orgs they don't own. We need to use Quartz' operator token to fetch those from IDPE.

Asalem1: Aww that makes sense.

src/unity/paths/me.yml
ischolten: Oh i was trying to make this all Accounts plural. It looks like there are lots of different tags and i was trying to make all the account(s) in sync. i can go through another iteration of unifying tags.

src/unity/paths/operator_accounts_accountId_users_userId.yml
ischolten: im not sure if was a standard, i can look at other contracts and adjust.

randycoulman (approved):
Changes LGTM!

PR #124, opened 2021-06-10
chore: add new properties for geo view

Palakp41:
Adds new properties for customizing s2/lat/lon columns onto the GeoViewProperties.
Adds new properties for tooltip columns onto GeoLayerProperties

### DIFF
--- a/src/common/schemas/GeoPointMapViewLayer.yml
+++ b/src/common/schemas/GeoPointMapViewLayer.yml
@@ -16,3 +16,6 @@
         isClustered:
           description: Cluster close markers together
           type: boolean
+        tooltipColumns:
+          description: An array for which columns to display in tooltip
+          type: array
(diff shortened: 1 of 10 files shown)


glinton (commented):
can you run `make generate` and push again

glinton (changes requested):
add an entry in `src/common/_schemas.yml` for `LatLonColumn` and `LatLonColumns` and re-generate, the oats check should then pass

src/common/schemas/GeoPointMapViewLayer.yml
```diff
@@ -16,3 +16,6 @@
         isClustered:
           description: Cluster close markers together
           type: boolean
+        tooltipColumns:
+          description: An array for which columns to display in tooltip
+          type: array
```
glinton, line 21: you may need to define the type for the items in this array

glinton, line 21:
```suggestion
          type: array
          items:
              type: string
```

glinton (approved):
:tada: nice work. don't consume this sha in the ui until your code is merged to cloud

PR #127, opened 2021-06-11
chore: updating some definitions that were getting compiled in funny ways

Asalem1:
Updating some schemas and definitions since oats was having a hard time


glinton (approved):
:sweet_potato:

PR #131, opened 2021-06-14
Remove redundant field from bucket metadata manifest

danxmoran:


src/oss/schemas/BucketMetadataManifest.yml
```diff
@@ -11,8 +11,6 @@ properties:
     type: string
   description:
     type: string
-  retentionRules:
```
danxmoran, line 14: The only reason to have this in addition to `retentionPolicies` below was to get easy access to the retention values in seconds. Now that we're using raw `time.Duration` it's easy to use `retentionPolicies` for recreating buckets.

wbaker85 (approved):

PR #133, opened 2021-06-15
Add backup/restore routes to OSS contract.

danxmoran:
Now that these are implemented on the server-side, add them to the API contract.


src/oss/paths/restore_bucket-metadata.yml
```diff
@@ -1,5 +1,5 @@
 post:
-  operationId: PostRestoreBucketMetadataId
+  operationId: PostRestoreBucketMetadata
```
danxmoran, line 2: Opportunistic change for something I missed when renaming the file earlier.

wbaker85 (approved):
:shipit:

PR #134, opened 2021-06-15
chore: add connection route

subirjolly:
Add a connection route so that we can provide it `email` parameter and retrieve its corresponding SSO connection.

### DIFF
--- a/contracts/priv/unity.yml
+++ b/contracts/priv/unity.yml
@@ -966,6 +966,34 @@ paths:
         default:
           description: Unexpected error
           $ref: '#/components/responses/ServerError'
+  /api/v2private/auth/connection:
+    get:
+      operationId: GetConnection
+      tags:
+        - Connection
+      parameters:
+        - $ref: '#/components/parameters/TraceSpan'
+        - in: query
+          name: email
+          schema:
+            type: string
+            format: email
+          required: true
+          description: The email to get connection for.
+      responses:
+        '200':
+          content:
+            text/plain:
+              schema:
+                type: string
+                example: connection
+          description: Connection name as string
+        '204':
+          description: No connection for the specified email
+        default:
+          description: Unexpected error
+          $ref: '#/components/responses/ServerError'
+      summary: Get connection  for the specified email
 components:
   parameters:
     TraceSpan:
--- a/contracts/priv/unity.yml
+++ b/contracts/priv/unity.yml
@@ -966,7 +966,7 @@ paths:
         default:
           description: Unexpected error
           $ref: '#/components/responses/ServerError'
-  /api/v2private/auth/connection:
+  /api/v2/quartz/auth/connection:
     get:
       operationId: GetConnection
       tags:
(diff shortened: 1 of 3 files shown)


src/unity.yml
```diff
@@ -61,6 +61,8 @@ paths:
     $ref: './unity/paths/operator_orgs_orgId.yml'
   '/operator/orgs/{orgId}/limits':
     $ref: './unity/paths/operator_orgs_orgId_limits.yml'
+  '/api/v2private/auth/connection':
```
Asalem1, line 64: this should be to the `api/v2/quartz` path

Asalem1 (changes requested):

src/unity.yml
Asalem1: Actually, sorry, this should just be `/auth/connection`

src/unity.yml
```diff
@@ -61,6 +61,8 @@ paths:
     $ref: './unity/paths/operator_orgs_orgId.yml'
   '/operator/orgs/{orgId}/limits':
     $ref: './unity/paths/operator_orgs_orgId_limits.yml'
+  '/api/v2/quartz/auth/connection':
```
Asalem1, line 64:
```suggestion
  '/auth/connection':
```

contracts/priv/unity.yml
```diff
@@ -966,6 +966,34 @@ paths:
         default:
           description: Unexpected error
           $ref: '#/components/responses/ServerError'
+  /api/v2/quartz/auth/connection:
+    get:
+      operationId: GetConnection
+      tags:
+        - Connection
+      parameters:
+        - $ref: '#/components/parameters/TraceSpan'
```
randycoulman, line 975: For my own enlightenment: What does this line do?

contracts/priv/unity.yml
```diff
@@ -966,6 +966,34 @@ paths:
...
+          name: email
+          schema:
+            type: string
+            format: email
+          required: true
+          description: The email to get connection for.
+      responses:
+        '200':
+          content:
+            text/plain:
```
randycoulman, line 986: I think we're going to return this as `application/json`; not sure if that will matter to the generated client code.

randycoulman (approved):
LGTM, though I think the response type might be wrong. I'll approve, but please double-check that before merging.

contracts/priv/unity.yml
subirjolly: I am not sure what TraceSpan does but I think we pass some meta info in headers through this.

contracts/priv/unity.yml
subirjolly: ah. what's the structure going to be for this response?

randycoulman: A JSON-encoded string. See https://github.com/influxdata/quartz/pull/4799 for an example curl request.

randycoulman:
I just remembered one more thing: this endpoint doesn't require any auth credentials; not sure if that needs to be called out somehow here.  We need to be able to make this call prior to a user being logged in, so there's no way to authenticate it using our usual mechanisms.

Asalem1 (approved):

contracts/priv/unity.yml
subirjolly: I don't think there's any way to specify if a route is authenticated.

PR #135, opened 2021-06-16
feat: orgIDs where needed for annotations & streams on OSS

wbaker85:
Related to https://github.com/influxdata/ui/issues/1645

For annotations (and streams) on OSS, we need to a way to know what the "active" orgID is for requests coming into the API for several paths. Since the concept of what the active org is is contained entirely in the frontend, this update includes orgID parameters that can be passed in the query where needed to support annotations on OSS.


scripts/reference.sh
```diff
@@ -20,6 +20,7 @@ echo "Aggregating oss swaggers"
 swagrag \
   -file ${CONTRACTS}/oss.yml \
   -file ${CONTRACTS}/mapsd.yml \
+  -file ${CONTRACTS}/priv/annotationd.yml \
```
wbaker85, line 23: Added this here, although I'm not 100% sure what impact it has.

glinton: this will include the annotations api in the published oss api, if the service is in, this is fine, otherwise i'd probably take this part out.

src/svc/annotationd/paths/annotations.yml
```diff
@@ -3,6 +3,11 @@ post:
     - Annotations
   operationId: createAnnotations
   summary: Create annotations
+  parameters:
+    - name: orgID
+      in: query
+      schema:
+        type: string
```
glinton, line 10: Since orgID as a query parameter is oss-only (annotations in cloud derives the org-id from the token), consider creating a `src/svc/annotationd-oss` that defines these custom paths and re-uses whatever it needs in terms of schemas.

glinton (changes requested):

glinton (approved):

PR #140, opened 2021-06-21
feat: add ping endpoint

bednar:
The users of our client libraries (`java`, `python`, `ruby`, ...) wants to do three things with our libraries: `query`, `write` and check the status of InfluxDB.

The query and writes are fine, but the checking status by `/health` endpoint brings trouble because the mapping is different for `oss` and `cloud` version. For more info see - https://github.com/influxdata/influxdb/issues/19357

As a solution this PR expose the `/ping` endpoint which works correctly in OSS and also in Cloud.

- https://github.com/influxdata/influxdb-client-python/issues/211
- https://github.com/influxdata/influxdb-client-ruby/issues/56
- https://github.com/influxdata/influxdb/issues/19357
- https://github.com/influxdata/influxdb/pull/21723

### DIFF
--- a/contracts/cloud.yml
+++ b/contracts/cloud.yml
@@ -59,7 +59,7 @@ paths:
     servers:
       - url: ''
     get:
-      operationId: PostPing
+      operationId: GetPing
       summary: Checks the status of InfluxDB instance and version of InfluxDB.
       responses:
         '204':
@@ -74,7 +74,7 @@ paths:
                 type: integer
               description: The version of InfluxDB.
     head:
-      operationId: PostPing
+      operationId: HeadPing
       summary: Checks the status of InfluxDB instance and version of InfluxDB.
       responses:
         '204':
--- a/contracts/cloud.yml
+++ b/contracts/cloud.yml
@@ -55,6 +55,39 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
+  /ping:
+    servers:
+      - url: ''
+    get:
+      operationId: PostPing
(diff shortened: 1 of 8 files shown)


contracts/cloud.yml
```diff
@@ -55,6 +55,39 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
+  /ping:
+    servers:
+      - url: ''
+    get:
+      operationId: PostPing
```
Asalem1, line 62: should this be a `post` or `get`?

src/common/paths/ping.yml
```diff
@@ -0,0 +1,30 @@
+get:
+  operationId: PostPing
```
glinton, line 2: Can you make these `operationId`s unique?

glinton (changes requested):

contracts/cloud.yml
bednar: It should be `GetPing`... fixed

src/common/paths/ping.yml
bednar: Fixed

src/common/paths/ping.yml
```diff
@@ -0,0 +1,30 @@
...
+    '204':
+      description: OK
+      headers:
+        X-Influxdb-Build:
+          schema:
+            type: string
+          description: The type of InfluxDB build.
+        X-Influxdb-Version:
+          schema:
+            type: integer
```
glinton, line 14: looks like this is actually a float, small enough nit i'll merge this

glinton (approved):

PR #141, opened 2021-06-21
fix: corrections for v1 dbrp api definition

wbaker85:
Related to https://github.com/influxdata/influx-cli/issues/24

This PR updates the swagger definition with the `v1dbrps` endpoint to more accurately reflect its usage, and allow for CLI code to be generated off of the swagger spec.

There were generally some mismatches between the form of responses and the types of parameters needed from requests that are being corrected with this PR.

### DIFF
--- a/src/common/paths/dbrps.yml
+++ b/src/common/paths/dbrps.yml
@@ -7,8 +7,12 @@ get:
     - $ref: "../parameters/TraceSpan.yml"
     - in: query
       name: orgID
-      required: true
       description: Specifies the organization ID to filter on
+      schema:
+        type: string
+    - in: query
+      name: org
+      description: Specifies the organization name to filter on
       schema:
         type: string
     - in: query
@@ -68,7 +72,7 @@ post:
     content:
       application/json:
         schema:
-          $ref: "../schemas/DBRP.yml"
+          $ref: "../schemas/DBRPCreate.yml"
   responses:
     "201":
       description: Database retention policy mapping created
--- a/src/common/paths/dbrps_dbrpID.yml
+++ b/src/common/paths/dbrps_dbrpID.yml
@@ -76,7 +76,7 @@ patch:
       content:
         application/json:
           schema:
-            $ref: "../schemas/DBRPGet.yml"
+            $ref: "../schemas/DBRPUpdateResponse.yml"
     "404":
       description: The mapping was not found
       content:
--- a/src/common/paths/dbrps_dbrpID.yml
+++ b/src/common/paths/dbrps_dbrpID.yml
@@ -76,7 +76,7 @@ patch:
       content:
         application/json:
           schema:
-            $ref: "../schemas/DBRPUpdateResponse.yml"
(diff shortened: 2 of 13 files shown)


src/common/paths/dbrps.yml
```diff
@@ -7,10 +7,14 @@ get:
     - $ref: "../parameters/TraceSpan.yml"
     - in: query
       name: orgID
-      required: true
```
wbaker85, line 10: Most endpoints allow specifying either the `orgID` or `org` (name), so the org name parameter was added in several places and the `orgID` required designation was removed. The "one or the other" requirement is handled in the application code.

src/common/paths/dbrps.yml
```diff
@@ -68,7 +72,7 @@ post:
     content:
       application/json:
         schema:
-          $ref: "../schemas/DBRP.yml"
+          $ref: "../schemas/DBRPCreate.yml"
```
wbaker85, line 75: Trying to use the same schema for the create body and the response was problematic, so a separate schema was created.

src/common/paths/dbrps_dbrpID.yml
```diff
@@ -23,7 +27,7 @@ get:
       content:
         application/json:
           schema:
-            $ref: "../schemas/DBRP.yml"
+            $ref: "../schemas/DBRPGet.yml"
```
wbaker85, line 30: `DBRPGet` is the response schema for the `get` method (and the `patch` method, see below) - which basically wraps the `DBRP` object in an object with the key `content`. This matches what the server produces.

src/common/schemas/DBRP.yml
```diff
@@ -24,14 +24,3 @@
       description: Specify if this mapping represents the default retention policy for the database specificed.
     links:
       $ref: "./Links.yml"
-  oneOf:
```
wbaker85, line 27: This part was breaking the CLI codegen. The `DBRP` schema now represents the DBRP response from the server, rather than trying to specify what the required inputs are as well as what the response looks like.

src/common/schemas/DBRPUpdate.yml
```diff
@@ -1,11 +1,6 @@
   properties:
-    database:
```
wbaker85, line 2: You can only update the retention policy and "default" designation through the update API, so the "links" and "database" fields are removed here.

src/common/schemas/DBRP.yml
danxmoran: Can we mark `bucketID`, `database`, and `retention_policy` as required to make up for this? It makes codegen nicer to have required things be marked.

Since `DBRP` is returned on GET requests, will `org` and `orgID` always be filled in? Could they also be marked as required?

src/common/schemas/DBRPGet.yml
```diff
@@ -0,0 +1,4 @@
+  type: object
+  properties:
+    content:
```
danxmoran, line 3: Can we mark this as required?

danxmoran (approved):

PR #143, opened 2021-06-21
chore: update response type

subirjolly:
The response type for sso connection api request needs to be application/json even though the response is actually a JSON encoded string. So, we aren't going to have a schema instead we will have the response as `application/json` and the schema type will be `string`.


randycoulman (approved):
LGTM 👍

PR #145, opened 2021-06-22
feat(operator-orgs): Update operator/orgs GET to reflect api response

hansenreid:
As part of [#4767](https://github.com/influxdata/quartz/issues/4767) I found that the current api response does not line up to the swagger definition. This updates the swagger to be in line with the response.

### DIFF
--- a/src/unity/schemas/RelatedAccount.yml
+++ b/src/unity/schemas/RelatedAccount.yml
@@ -21,4 +21,4 @@ properties:
   updatedAt:
     type: string
     description: date of last update to account
-required: [id, balance, type, billingContact]
+required: [id, balance, type, deletable, updatedAt]
--- a/src/unity/schemas/RelatedAccount.yml
+++ b/src/unity/schemas/RelatedAccount.yml
@@ -21,4 +21,4 @@ properties:
   updatedAt:
     type: string
     description: date of last update to account
-required: [id, balance, type, deletable, updatedAt]
+required: [id, balance, type, deletable, updatedAt, billingContact]
--- a/src/unity/schemas/RelatedAccount.yml
+++ b/src/unity/schemas/RelatedAccount.yml
@@ -2,14 +2,23 @@ description: Subset of Account data related to the organization
 type: object
 properties:
   id:
-    type: string
-    description: account id
-  email:
-    type: string
-    description: email associated with the account
+    type: number
+    description: account id in quartz
   type:
     $ref: './AccountType.yml'
+  zuoraAccountId:
+    type: string
+    description: Zuora ID associated with the account
+  deletable:
+    type: boolean
+    description: flag whether the account can be deleted or not
   balance:
     type: number
     description: remaining balance on the account, nil if none
-required: [id, email, type, balance]
+  billingContact:
+    $ref: './BillingContact.yml'
+    description: billing contact for the account
+  updatedAt:
+    type: string
+    description: date of last update to account
+required: [id, balance, type, billingContact]
--- a/src/unity/schemas/RelatedAccount.yml
+++ b/src/unity/schemas/RelatedAccount.yml
@@ -6,12 +6,6 @@ properties:
     description: account id in quartz
   type:
     $ref: './AccountType.yml'
-  zuoraAccountId:
-    type: string
-    description: Zuora ID associated with the account
-  deletable:
-    type: boolean
-    description: flag whether the account can be deleted or not
   balance:
     type: number
     description: remaining balance on the account, nil if none
@@ -21,4 +15,4 @@ properties:
   updatedAt:
     type: string
     description: date of last update to account
-required: [id, balance, type, deletable, updatedAt, billingContact]
+required: [id, balance, type, updatedAt]
--- a/contracts/priv/unity.yml
+++ b/contracts/priv/unity.yml
@@ -1574,12 +1574,12 @@ components:
         balance:
           type: number
           description: 'remaining balance on the account, nil if none'
-        updatedAt:
-          type: string
-          description: date of last update to account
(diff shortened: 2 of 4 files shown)


Asalem1 (changes requested):
I think this might need to be updated in the `unity/*` relevant files and autogenerated when running the `make generate-all` command

gspletzer:
> I think this might need to be updated in the `unity/*` relevant files and autogenerated when running the `make generate-all` command

That is correct. Both contracts and the `unity/` files need to be updated.

src/unity/schemas/RelatedAccount.yml
```diff
@@ -2,14 +2,23 @@ description: Subset of Account data related to the organization
 type: object
 properties:
   id:
-    type: string
-    description: account id
-  email:
```
Asalem1, line 7: we still need the email here to display the organization's account email in the operator/orgs page, unless that's accessible via another property like the billing contact? But if that's the case then we'd need to update the billingContact in the RelatedAccount type to be required rather than optional. Thoughts?

Asalem1 (approved):
I think the only sticking point for me is removing the `email` from the `RelatedAccount` type. I don't think the `zuoraAccountId` or the `deletable` properties are being used in the UI for the `RelatedAccount` types (although they are being used for the base `Account` type) so I don't think those need to be included here.

src/unity/schemas/RelatedAccount.yml
gspletzer: Email is currently coming in with the billing contact. The reason I made billing contact not required is because both `operator/orgs` and `operator/orgs/:orgId` share the related account schema, but the latter doesn't currently return billing contact. I suppose we could keep it as required and it can just be an empty object for the latter?

Asalem1: making the billingContact required sounds good to me

gspletzer:
> I think the only sticking point for me is removing the `email` from the `RelatedAccount` type. I don't think the `zuoraAccountId` or the `deletable` properties are being used in the UI for the `RelatedAccount` types (although they are being used for the base `Account` type) so I don't think those need to be included here.

I believe these were added here because that is the shape the api is currently returning.

src/unity/schemas/RelatedAccount.yml
gspletzer: Okay. I will add it back in. 😄

Asalem1:
> > I think the only sticking point for me is removing the `email` from the `RelatedAccount` type. I don't think the `zuoraAccountId` or the `deletable` properties are being used in the UI for the `RelatedAccount` types (although they are being used for the base `Account` type) so I don't think those need to be included here.
> 
> I believe these were added here because that is the shape the api is currently returning.

gotcha. As far the UI is concerned it's just noise but if it requires less work from the API then that's fine

gspletzer:
@Asalem1 I've pushed up the change to make billing contact a requirement. I assume if it has no value, it will just appear empty? Currently there is no billing contact property being returned for the `operator/orgs/:orgId`, so just want to make sure we won't be breaking anything since it is required.

Asalem1:
> @Asalem1 I've pushed up the change to make billing contact a requirement. I assume if it has no value, it will just appear empty? Currently there is no billing contact property being returned for the `operator/orgs/:orgId`, so just want to make sure we won't be breaking anything since it is required.

i'll check in a sec

randycoulman:
If there are fields that the UI doesn't need, IMO it's OK to not document them in the schema. Quartz can return extra fields if it wants to and the UI can ignore them (referencing the conversation above about `zuoraAccountId` and `deletable`.

Asalem1:
> @Asalem1 I've pushed up the change to make billing contact a requirement. I assume if it has no value, it will just appear empty? Currently there is no billing contact property being returned for the `operator/orgs/:orgId`, so just want to make sure we won't be breaking anything since it is required.

Looks good to me

src/unity/schemas/RelatedAccount.yml
randycoulman: What impact does marking a field `required` have on the client code? My initial assumption was that it's really more for POST/PUT/PATCH requests and not GETs. Are there more semantics to it than that?

I don't actually care one way or the other whether we mark billing contact as required, but I don't want to have to return it from the specific org GET request if it isn't needed there because that's just increasing payload size (and database work) unnecessarily.  We know it's needed from the GET all orgs request, and it is returned there as needed.

So either way, the system isn't going to break.

So what's the best way to specify this distinction in the swagger without having to create another entity type? I have very limited experience with swagger, so I'm not sure what the right answer is here.

If we can mark the field as required without returning it from the one request, I'm fine with that. If we can mark the field as optional while still returning it from the other request, I'm fine with that as well. What does the client need from the swagger spec in order for this to work?

Asalem1: It's just to enforce the contract between the expected output for the UI. If having an email field for an associated account on the operator page is an optional field, then we can make the billing contact or email optional and go from there. I don't have much context on the usage of the operator page so it's hard for me to make a judgment call on what should exist and what shouldn't. Based on the UI that was ported over, the expectation was that an associated account email would exist for the org/id path, but if that's not the case then we can just mark it as option and leave it blank if none exists

gspletzer: Final decision: `zuoraAccountId` and `deletable` have been removed from `RelatedAccount` schema; they may still be sent, but aren't part of the swagger contract. `billingContact` has been made optional since it is only needed for one of the routes that utilize the schema, as discussed/approved by @Asalem1 .

Asalem1 (approved):

PR #147, opened 2021-06-23
feat: add endpoint for v1 auth set password

DStrand1:
Adds an endpoint for [this method here](https://github.com/influxdata/influxdb/blob/master/v1/authorization/http_server.go#L636-L664), which is needed for the InfluxCLI `influx v1 auth` command


src/oss.yml
```diff
@@ -24,6 +24,8 @@ paths:
     $ref: "./common/paths/authorizations.yml"
   /authorizations/{authID}:
     $ref: "./common/paths/authorizations_authID.yml"
+  /authorizations/{authID}/password:
```
danxmoran, line 27: This needs a `servers` to set the `/private/legacy` prefix, yes? Otherwise it'll fall under the default `/api/v2` prefix.

src/oss.yml
```diff
@@ -24,6 +24,8 @@ paths:
     $ref: "./common/paths/authorizations.yml"
   /authorizations/{authID}:
     $ref: "./common/paths/authorizations_authID.yml"
```
danxmoran, line 26: The `/authorizations` routes above here are for v2 auth. We need to add copies with `servers` set to also cover the v1 endpoints.

DStrand1:
I think the change to `ref/oss.yml` is not correct, since it is put as `api/v2` currently

contracts/oss.yml
```diff
@@ -5590,6 +5590,42 @@ paths:
         default:
           description: Unexpected error
           $ref: '#/components/responses/ServerError'
+  /legacy/authorizations:
+    servers:
+      - url: /private
+    $ref: '#/paths/~1authorizations'
```
danxmoran, line 5596: I'm surprised these are showing up in the "normal" `oss.yml` contract... I think I know a way to fix it, but it'll require some copy-paste.

src/oss.yml
```diff
@@ -24,6 +24,18 @@ paths:
...
+  /legacy/authorizations:
+    servers:
+      - url: "/private"
+    $ref: "./common/paths/authorizations.yml"
```
danxmoran, line 30: I think to avoid the weird `$ref` values in the contracts, we'll need to stop reusing files between the v1 and v2 operations. I'd try:
1. Copy this file to `oss/paths/legacy_authorizations.yml`
2. Change the tag in the file from its current value (I assume `Authorizations`) to something like `Legacy Authorizations`
3. Update this ref to point at the new file

Then I'd do the same with the `authorizations_authID.yml` file below

src/oss/paths/authorizations_authID_password.yml
```diff
@@ -0,0 +1,25 @@
+patch:
```
danxmoran, line 1: I'd rename this file to `legacy_authorizations_authID_password.yml` to match the suggestions above

src/oss/paths/authorizations_authID_password.yml
```diff
@@ -0,0 +1,25 @@
...
+  operationId: SetAuthorizationsIDPassword
+  tags:
+    - Authorizations
```
danxmoran, line 4:
```suggestion
    - Legacy Authorizations
```

danxmoran (approved):

PR #150, opened 2021-06-24
feat: add token param to get v1 auth ID

DStrand1:
Adds the token param for v1 auth get by ID path


src/oss/paths/legacy_authorizations_authID.yml
```diff
@@ -5,12 +5,16 @@ get:
   summary: Retrieve an authorization
   parameters:
     - $ref: "../../common/parameters/TraceSpan.yml"
-    - in: path
+    - in: query
```
danxmoran, line 8: I think this change also requires updating the path in `src/oss.yml` so it doesn't have the trailing `/{authID}`

danxmoran (approved):

PR #157, opened 2021-06-28
feat: add response code to managed functions invoke by id

wbaker85:
Related to https://github.com/influxdata/influx-cli/issues/153

Adds a `200` response code designation to the `functions/{functionID}/invoke` path, with a default to being a server error.

This is being done to allow the CLI codegen to work properly for this path.


yquansah:
I will make more improvements to this swagger definition

yquansah (approved):

PR #160, opened 2021-06-30
feat(unity): Add route for `DELETE /account`

gspletzer:
Part of [Quartz #4360](https://github.com/influxdata/quartz/pull/4859)

Adds definition for `DELETE /account` to Unity swagger. This API will allow a user to self-delete an account, as long as the account is `free` and there are no other users.

### DIFF
--- a/contracts/priv/unity.yml
+++ b/contracts/priv/unity.yml
@@ -29,6 +29,27 @@ paths:
         default:
           description: Unexpected error
           $ref: '#/components/responses/ServerError'
+  /account:
+    delete:
+      operationId: DeleteAccount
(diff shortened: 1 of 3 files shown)


Asalem1:
I think I remember this being on the `users` page right? If that's the case, shouldn't this need to be `orgs_orgId_account`?

gspletzer:
> I think I remember this being on the `users` page right? If that's the case, shouldn't this need to be `orgs_orgId_account`?

Not sure. I just added a conversation to UI unification channel on slack so the group can discuss.

Asalem1 (approved):
LGTM.

PR #167, opened 2021-07-16
feat: add some missing pieces and consolidate the source of truth

Asalem1:
This PR adds a few definitions for the UI to replace some existing UI-based sources of truth with a centralized yaml definition. Who'd of thought. I'd be down to go through each of these to verify whether they are in fact true in their assertions with the API.


src/cloud/paths/experimental_sampledata_buckets_members.yml
```diff
@@ -0,0 +1,44 @@
...
+    - in: path
+      name: bucketID
+      schema:
+        type: string
+  responses:
+    '204':
+      description: A list of demo data buckets
+    '200':
+      description: if sampledata route is not available gateway responds with 200
+    default:
```
Asalem1, line 17: This was weird but it doesn't look like the UI cares about data being returned besides a status code

src/cloud/paths/experimental_sampledata_buckets_members.yml
```diff
@@ -0,0 +1,44 @@
...
```
Asalem1, line 39: same as above

src/cloud/paths/oauth_clientConfig.yml
```diff
@@ -2,13 +2,20 @@ get:
   tags:
     - OAuth
   summary: Retrieve required OAuth client variables
+  parameters:
+    - in: query
+      name: redirectTo
+      schema:
+        type: string
+      description: optional parameter to redirect the user to after a successful login.
+      required: false
```
Asalem1, line 11: we'll see if this works  🤞🏽

src/cloud/paths/experimental_sampledata_buckets_members.yml
glinton: that is interesting.. it returns `no content` when there's a list of buckets, and an `ok` when the route does **not** exist (and no content is returned otherwise)?

src/cloud/paths/experimental_sampledata_buckets_members.yml
glinton: :man_shrugging:

glinton (approved):

src/cloud/paths/experimental_sampledata_buckets_members.yml
Asalem1: This is just an assumption, but I think that the feature intent here is that we add a member to a bucket whenever they try and access the demo data bucket, and so we just make a request to add or remove the user from the bucket without actually caring what the outcome is since we don't actually care to return a list of members. I _think_ the purpose is to enable the demodata bucket for a user? 🤷🏽

PR #168, opened 2021-07-22
feat: openapi routes for pinned items

ShmuelLotman:
As per description


glinton (commented):
There's no generated contract from this, are you planning on consuming this api?

glinton (approved):

PR #171, opened 2021-07-30
feat: add operatorRole kv for me route

subirjolly:
closes: #2187

### DIFF
--- a/contracts/priv/unity.yml
+++ b/contracts/priv/unity.yml
@@ -1526,6 +1526,9 @@ components:
         isOperator:
           type: boolean
           description: whether the user is an operator
+        operatorRole:
+          type: string
+          description: role permission whether this operator is "read-only" | "read-write"
       required:
         - id
         - accountType
--- a/src/unity/schemas/Me.yml
+++ b/src/unity/schemas/Me.yml
@@ -21,4 +21,7 @@ properties:
   isOperator:
     type: boolean
     description: whether the user is an operator
+  operatorRole:
+    type: string
+    description: role permission whether this operator is "read-only" | "read-write"
 required: [id, accountType, billingProvider, email, isOperator, isRegionBeta]


contracts/priv/unity.yml
```diff
@@ -1526,6 +1526,9 @@ components:
         isOperator:
           type: boolean
           description: whether the user is an operator
+        operatorRole:
+          type: string
```
Asalem1, line 1530: can we make this into an enum?

Asalem1 (commented):
Is there a link to the issue? Does the API support this?

joelfried (approved):

contracts/priv/unity.yml
subirjolly: good call. i made it an interface(enum like) in quartz. missed here. doing that.

subirjolly: @Asalem1 done!

subirjolly:
@Asalem1 Here's the epic: https://app.zenhub.com/workspaces/applicationsui-unity-606b1041215d5f0017898847/issues/influxdata/ui/2174

The API doesn't support it yet but very soon it will. I will merge it after talking to Reid.

Asalem1 (approved):

hansenreid:
Should the field be marked as `nullable: true` since non operators will have a null value for operatorRole?

Also, since the backend work for this won't happen until next sprint, will adding the field beforehand break anything on your end?

subirjolly:
@Asalem1 @hansenreid Made it nullable.

hansenreid (approved):

Asalem1 (approved):

subirjolly:
> Should the field be marked as `nullable: true` since non operators will have a null value for operatorRole?
> 
> Also, since the backend work for this won't happen until next sprint, will adding the field beforehand break anything on your end?

I just finished testing with OpenAPI and UI having `operatorRole` and quartz-mock without it. It seems fine. But if your code changes will be out next week, then I'd hold off on merging this as I am out tomorrow and don't want broken stuff(Just in Case) to be left in master.

PR #174, opened 2021-08-11
refactor(unity): inline and eliminate Account schema

randycoulman:
We are about to add a new Quartz API for switching between multiple accounts. For that, we'd like to use the `Account` schema, but the shape is going to be different than what's here.

The `Account` schema is currently only used within the `OperatorAccount` schema, so this change inlines `Account` into `OperatorAccount` and removes both `Account` and the unused `Accounts` schemas.

A later PR will reintroduce those schemas in a new form for the account-switching API.

While working on this, I noticed that we hadn't added the new `contract` enum value to `AccountType`, so I've added it here.


Asalem1 (approved):
LGTM. Doesn't seem like anything will break on the UI as a result of this

PR #176, opened 2021-08-12
feat(unity): add account-switching API

randycoulman:
Adds APIs for getting a list of the current user's accounts and for allowing the user to select a different default account.

These endpoints won't be implemented in Quartz for a few sprints, but having them defined now will allow the UI team to mock them out and start building out the UI for this feature.

When the implementation starts in Quartz, we'll also move the `DELETE /account` endpoint to `DELETE /accounts/:id` for consistency. See https://github.com/influxdata/quartz/issues/5077.

### DIFF
--- a/contracts/priv/unity.yml
+++ b/contracts/priv/unity.yml
@@ -50,6 +50,67 @@ paths:
         default:
           description: Unexpected error
           $ref: '#/components/responses/ServerError'
+  /accounts:
+    get:
+      operationId: GetAccounts
+      tags:
+        - Accounts
+      summary: Get the list of accounts for the current user
+      responses:
+        '200':
+          description: Accounts
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  properties:
+                    id:
+                      type: number
+                      description: account id in quartz
+                    isActive:
+                      type: boolean
+                      description: is this the currently active account in the session?
+                    isDefault:
+                      type: boolean
+                      description: is this the user's default account?
+                    name:
+                      type: string
+                      description: name of the account
+                  required:
+                    - id
+                    - isActive
+                    - isDefault
+                    - name
+        '401':
+          description: Unauthorized
+          $ref: '#/components/responses/ServerError'
+        default:
+          description: Unexpected error
+          $ref: '#/components/responses/ServerError'
+  /accounts/default:
+    put:
+      operationId: PutDefaultAccount
+      tags:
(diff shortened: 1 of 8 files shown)


lexun (approved):

jrenee42 (commented):
i don't see the renaming of the delete action; from the slack convo with @randycoulman 

Rename DELETE /account -> DELETE /accounts/:id
also; you wrote "isCurrent" and you changed it to 'isActive".  either is fine, just curious why it was switched.  thanks!
https://influxdata.slack.com/archives/C0222GPT5FY/p1628780568015700?thread_ts=1628779255.005500&cid=C0222GPT5FY

randycoulman:
I wrote about the `DELETE` rename in the PR description:

> When the implementation starts in Quartz, we'll also move the DELETE /account endpoint to DELETE /accounts/:id for consistency. See influxdata/quartz#5077.

I didn't want to change the swagger for it yet because it would affect currently-deployed code and the Quartz backend for that change isn't in place yet.

Regarding `isCurrent` vs `isActive`, I went with `isActive` because you suggested it as a better name.

contracts/priv/unity.yml
```diff
@@ -50,6 +50,67 @@ paths:
...
+          $ref: '#/components/responses/ServerError'
+        default:
+          description: Unexpected error
+          $ref: '#/components/responses/ServerError'
+  /accounts/default:
+    put:
+      operationId: PutDefaultAccount
+      tags:
+        - Accounts
+      requestBody:
```
hoorayimhelping, line 96: out of curiosity, why put the id in a request body and not as part of the url? e.g.

```
accounts/default/account_id
```

hoorayimhelping (approved):
works for me, pretty straightforward.

thanks for putting this out there for us to work against ahead of it being implemented! good thinkin'

contracts/priv/unity.yml
randycoulman: No real reason other than it looked kinda weird to me. I also considered `accounts/:account_id/default`, but that also felt weird.

If putting it the URL is better for you, I'm happy to change it. Note, though, that because we enforce JSON content type, you'll have to pass an empty body with that.

Let me know your preference and I'll either update or merge.

hoorayimhelping: no preference, just curious. I think enforcing a JSON content type and needing a body means this is fine :)

PR #178, opened 2021-08-16
feat: add OSS APIs for managing remote connections

danxmoran:
Closes influxdata/influxdb#22202

Open-ish questions:

- [ ] Can we think of a better prefix than `/remotes`?
- [ ] Do we need to add self-links / pagination links to responses?

### DIFF
--- a/contracts/oss-diff.yml
+++ b/contracts/oss-diff.yml
@@ -1808,6 +1808,175 @@ paths:
         default:
           description: Unexpected error
           $ref: '#/components/responses/ServerError'
+  /remotes:
+    get:
+      operationId: GetRemoteConnections
+      tags:
+        - RemoteConnections
+      summary: List all remote connections
+      parameters:
+        - $ref: '#/paths/~1ready/get/parameters/0'
+        - in: query
+          name: org
+          schema:
+            type: string
+        - in: query
+          name: orgID
+          description: The organization ID.
+          schema:
+            type: string
+        - in: query
+          name: name
+          schema:
+            type: string
+        - in: query
+          name: id
+          schema:
+            type: string
+        - in: query
+          name: remoteOrgId
+          schema:
+            type: string
+        - in: query
+          name: remoteUrl
+          schema:
+            type: string
+            format: uri
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RemoteConnections'
+        '404':
+          $ref: '#/components/responses/ServerError'
+        default:
+          $ref: '#/components/responses/ServerError'
+    post:
+      operationId: PostRemoteConnections
+      tags:
+        - RemoteConnections
+      summary: Register a new remote connection
+      parameters:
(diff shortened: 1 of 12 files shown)


davidby-influx (commented):
LGTM, but you need someone more experienced to review.

src/oss/schemas/RemoteConnection.yml
```diff
@@ -0,0 +1,25 @@
+type: object
```
wbaker85, line 1: Are these going to have remote bucket IDs or something like that associated with them?

src/oss/schemas/RemoteConnections.yml
```diff
@@ -0,0 +1,7 @@
+type: object
+properties:
+  # TODO(danmoran): Add pagination / links?
```
wbaker85, line 3: None of our other APIs support pagination that I know of so I think it's fine to leave this as a TODO for now. Pagination support here seems like a lower priority than it would be in some other areas as well.

wbaker85:
As far as the prefix...are these remotes intended to always be InfluxDB Cloud remotes, or could they possibly be something else like another OSS instance? Not sure if that is going to give me inspiration for other prefix names or not but it might narrow it down a bit.

src/oss/schemas/RemoteConnection.yml
danxmoran: There will be a separate set of API objects for write-replications, linking source bucket + target remote + target bucket + any other configs. The idea being that:
* There might be replication streams for many buckets, and it'd be annoying to enter the connection info repeatedly (especially when updating an auth token)
* In the future we might want to add more kinds of replications, which could reuse the remote-connection metadata

contracts/oss.yml
```diff
@@ -6964,6 +6964,177 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RemoteConnections'
+        '404':
+          $ref: '#/components/responses/ServerError'
+        default:
+          $ref: '#/components/responses/ServerError'
+    post:
+      operationId: PostRemoteConnections
```
wbaker85, line 7012: Should this be singular?  Seems like it's only for making one at a time.

danxmoran:
> are these remotes intended to always be InfluxDB Cloud remotes, or could they possibly be something else like another OSS instance?

AFAIK the primary product goal is OSS->Cloud, but we won't be doing anything that prevents OSS->OSS from working. @russorat can you weigh in?

contracts/oss-diff.yml
```diff
@@ -2650,6 +2821,82 @@ components:
...
         - oldId
         - newId
+    RemoteConnection:
+      type: object
+      properties:
+        id:
+          type: string
+        name:
+          type: string
+        orgId:
```
danxmoran, line 2831: Self-review: seems other APIs typically use `orgID`, would be good to be consistent

wbaker85 (approved):
Regardless of whether these remotes are cloud-exclusive or not, I can't really think of a better name for the resources than `remotes`, so LGTM!

PR #182, opened 2021-08-20
feat: add OSS APIs for managing replications

danxmoran:
Closes influxdata/influxdb#22163

### DIFF
--- a/src/oss/schemas/Replication.yml
+++ b/src/oss/schemas/Replication.yml
@@ -1,0 +1,32 @@
+type: object
+properties:
+  id:
+    type: string
+  name:
+    type: string
+  description:
+    type: string
+  remoteID:
+    type: string
+  localBucketID:
+    type: string
+  remoteBucketID:
+    type: string
+  maxQueueSizeBytes:
+    type: int
+    format: int64
+  currentQueueSizeBytes:
+    type: int
+    format: int64
+  latestResponseCode:
+    type: int
+  latestErrorMessage:
+    type: string
+required:
+  - id
+  - name
+  - remoteID
+  - localBucketID
+  - remoteBucketID
+  - maxQueueSizeBytes
+  - currentQueueSizeBytes
--- a/src/oss/schemas/ReplicationCreationRequest.yml
+++ b/src/oss/schemas/ReplicationCreationRequest.yml
@@ -1,0 +1,26 @@
+type: object
+properties:
+  name:
+    type: string
+  description:
+    type: string
+  orgID:
+    type: string
+  remoteID:
+    type: string
+  localBucketID:
+    type: string
+  remoteBucketID:
+    type: string
+  maxQueueSizeBytes:
+    type: int
+    format: int64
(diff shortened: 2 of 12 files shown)


src/oss/schemas/ReplicationCreationRequest.yml
```diff
@@ -0,0 +1,26 @@
...
+  remoteID:
+    type: string
+  localBucketID:
+    type: string
+  remoteBucketID:
+    type: string
+  maxQueueSizeBytes:
+    type: int
+    format: int64
+    min: 33554430 # 32 MiB
```
wbaker85, line 18: Is there any significance to these numbers right now? I see in the linked issue it mentions a 10mb segment size.

src/oss/schemas/Replication.yml
```diff
@@ -0,0 +1,32 @@
...
+    type: int
+    format: int64
+  currentQueueSizeBytes:
+    type: int
+    format: int64
+  latestResponseCode:
+    type: int
+  latestErrorMessage:
+    type: string
+required:
```
wbaker85, line 25: Does this need a property for "skip-verify/insecure TLS"?

src/oss/schemas/ReplicationCreationRequest.yml
```diff
@@ -0,0 +1,26 @@
...
+    default: 67108860 # 64 MiB
+required:
```
wbaker85, line 20: Similar question for the "skip-verify/insecure TLS" on the `Replication` object.

src/oss/schemas/ReplicationCreationRequest.yml
danxmoran: Yes-ish. Our durable queue impl is written such that data is only reclaimed in increments of the segment-size (10mb by default). If the max queue size is "too small" you can get into a bad UX where already-read data in the head segment eats up so much of the queue's allowed space that writes start to fail. For example, if the max size was == the segment size, you could end up in a situation like:
1. 9.99mb of data is written to the queue
2. All the data is read from the queue, but not reclaimed because the segment isn't full yet
3. All writes bigger than 0.01mb are rejected because they'd push the queue's total disk usage over the max of 10mb
4. Write-rejections continue until enough small writes fill up the 0.01mb remaining in the segment

Sam and I decided that setting a minimum value for the max-queue-size would be safer than trying to edit the (old, very stable) durable queue impl. I went with the smallest power of 2 that was larger than 2x the segment size.

src/oss/schemas/Replication.yml
danxmoran: Nope, that's captured in the `RemoteConnection` referenced by `remoteID`

src/oss/schemas/ReplicationCreationRequest.yml
danxmoran: Same as above, captured in the referenced `remoteID`

wbaker85 (approved):

PR #183, opened 2021-08-24
fix: keyword for minimal value

bednar:
The keyword for minimal value is `minimum` - https://swagger.io/docs/specification/data-models/data-types/#range

image

This PR fix our generator - https://app.circleci.com/pipelines/github/bonitoo-io/influxdb-clients-apigen/151/workflows/98354d0f-d6fe-4986-be38-c972815473a8


glinton (commented):
These appear to be the same changes as those contained in #184

bednar:
Yes, it is duplicate of #184

PR #193, opened 2021-08-30
feature: improve and update write response descriptions. Edits are ba…

jstirnaman:
## Improvements for API reference docs
Screenshots
image
image

Improve error responses (based on testing).
Extract common error responses.
Add links to docs, docs URL placeholders, and scripts for placeholder substitution.
Add utility script for extracting tags.
Add sorted tags, tag groups, x-traitTags for custom content, and descriptions.
Add TokenAuth, QuerystringAuth, and documentation.

Along with Redoc config changes in 2f14d045b, displays grouped tags and moves Authentication to customizable section.

Related to https://github.com/influxdata/docs-v2/issues/3003. 
Related to https://github.com/influxdata/docs-v2/issues/2685.
Opened related issue https://github.com/influxdata/idpe/issues/11615 to make a 413 return JSON instead of HTML.

### DIFF
--- a/contracts/cloud-diff.yml
+++ b/contracts/cloud-diff.yml
@@ -20,7 +20,7 @@ tags:
       Retrieve data, analyze queries, and get query suggestions.
   - name: Write
     description: |
-      Write time-series data to buckets.
+      Write time series data to buckets.
   - name: Authorizations
     description: |
       Create and manage API tokens. An **authorization** associates a list of permissions to an **organization** and provides a token for API access. To assign a token to a specific user, scope the authorization to the user ID.
--- a/contracts/cloud-diff.yml
+++ b/contracts/cloud-diff.yml
@@ -2,8 +2,83 @@ openapi: 3.0.0
 info:
   title: Influx Cloud API Service
   version: 2.0.1
+  description: |
+    The InfluxDB v2 API provides a programmatic interface for all interactions with InfluxDB. Access the InfluxDB API using the `/api/v2/` endpoint.
 servers:
   - url: /api/v2
+tags:
+  - name: Authentication
+    description: |
+      Use one of the following schemes to authenticate to the InfluxDB API:
+      - [Token authentication](#section/Authentication/TokenAuth)
+      - [Basic authentication](#section/Authentication/BasicAuth)
+      - [Querystring authentication](#section/Authentication/QuerystringAuth)
+      <!-- ReDoc-Inject: <security-definitions> -->
+    x-traitTag: true
+  - name: Query
+    description: |
+      Retrieve data, analyze queries, and get query suggestions.
+  - name: Write
+    description: |
+      Write time-series data to buckets.
+  - name: Authorizations
+    description: |
+      Create and manage API tokens. An **authorization** associates a list of permissions to an **organization** and provides a token for API access. To assign a token to a specific user, scope the authorization to the user ID.
+x-tagGroups:
+  - name: Get started
+    tags:
+      - Authentication
+  - name: ' '
+    tags:
+      - Write
+      - Query
+  - name: Resource endpoints
+    tags:
+      - Buckets
+      - Dashboards
+      - Tasks
+  - name: Security and access endpoints
+    tags:
+      - Authorizations
+      - Organizations
+      - Users
+  - name: System information endpoints
+    tags:
+      - Ping
+      - Routes
+  - name: All endpoints
+    tags:
+      - Authorizations
+      - Buckets
+      - Cells
+      - Checks
+      - DBRPs
+      - Dashboards
+      - Delete
+      - DemoDataBuckets
+      - Labels
+      - Limits
+      - NotificationEndpoints
+      - NotificationRules
+      - OAuth
+      - Organizations
+      - Ping
+      - Public Flags
+      - Query
+      - Routes
+      - Rules
+      - Secrets
+      - Setup
+      - Signin
+      - Signout
+      - Sources
+      - Tasks
+      - Telegrafs
+      - Templates
+      - Usage
+      - Users
+      - Variables
+      - Write
 paths:
   /users:
     get:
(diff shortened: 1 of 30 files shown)


contracts/cloud.yml
```diff
@@ -1198,41 +1198,45 @@ paths:
             $ref: '#/components/schemas/WritePrecision'
       responses:
         '204':
-          description: Write data is correctly formatted and accepted for writing to the bucket.
+          description: 'The sent data was correctly formatted and accepted for writing to the bucket. `204` doesn''t indicate a successful write operation since writes are asynchronous. See [how to check for write errors](###some-link###).'
```
jstirnaman, line 1201: Waiting on the docs.influxdata.com content for the feature described in https://github.com/influxdata/docs-v2/issues/3003

contracts/cloud.yml
```diff
@@ -1198,41 +1198,45 @@ paths:
...
-          description: No token was sent and they are required.
+        '404':
+          description: 'Not found. A requested resource was not found. The response body contains the requested resource type, e.g. `organization name` or `bucket`, and name.'
           content:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
         '413':
-          description: Write has been rejected because the payload is too large. Error message returns max size supported. All data in body was rejected and not written.
+          description: Request entity too large. The payload exceeded the 50MB limit. All payload data was rejected and not written.
```
jstirnaman, line 1225: `413` is handled by Nginx before it hits the application. Nginx is returning HTML instead of JSON as specified. I opened https://github.com/influxdata/idpe/issues/11615 to correct that. In the meantime, should we remove or correct the content spec here?

jstirnaman:
This is in draft for now. See my comments for pending items.

contracts/cloud.yml
glinton: Changes to the openapi spec are quick enough it seems like in this case especially (discussed fix has no eta), we can go ahead and update the spec to match the api's behavior.

contracts/cloud.yml
glinton: I agree with this comment - https://github.com/influxdata/docs-v2/issues/3003#issuecomment-906787335, `202` feels more appropriate to add as a response, until then, this description update suffices.

contracts/cloud.yml
danxmoran: I believe OSS returns this error case as JSON, so :+1: to adding an HTML response case but I'd leave the JSON case as well.

jstirnaman: Does this limit exist in OSS? I haven't been able to hit one and I don't see a config property for it.

danxmoran: There is a code path in OSS to detect a too-big batch and return the correct error code, but the detection depends on Go's built-in `ioutil.ReadAll` raising an error and I'm not sure what the upper limit is within that (or if there even is one).

contracts/common.yml
```diff
@@ -1198,9 +1199,9 @@ paths:
             $ref: '#/components/schemas/WritePrecision'
       responses:
         '204':
-          description: 'The sent data was correctly formatted and accepted for writing to the bucket. `204` doesn''t indicate a successful write operation since writes are asynchronous. See [how to check for write errors](###some-link###).'
+          description: 'InfluxDB validated the request data format and accepted the data for writing to the bucket. `204` doesn''t indicate a successful write operation since writes are asynchronous. See [how to check for write errors]({{% INFLUXDB_DOCS_URL %}}/write-data/troubleshoot).'
```
jstirnaman, line 1202: @glinton ...and templating here.

scripts/generate.sh
```diff
@@ -1,12 +1,15 @@
...
 # generate oss contract
 sed -e '/#REF_COMMON_PATHS/{r ./src/common/_paths.yml' -e 'd}' src/oss.yml |
 sed -e '/#REF_COMMON_PARAMETERS/{r ./src/common/_parameters.yml' -e 'd}' |
 sed -e '/#REF_COMMON_SCHEMAS/{r ./src/common/_schemas.yml' -e 'd}' > src/.oss_gen.yml && \
-swagger-cli bundle src/.oss_gen.yml --outfile ${CONTRACTS}/oss.yml --type yaml && \
+swagger-cli bundle src/.oss_gen.yml --type yaml | \
+(sed -e "s|{{% INFLUXDB_DOCS_URL %}}|${INFLUXDB_DOCS_URL}/v2.0|" > ${CONTRACTS}/oss.yml)  && \
+swagger-cli bundle ${CONTRACTS}/oss.yml --outfile ${CONTRACTS}/oss.yml --type yaml
 swagger-cli bundle src/.oss_gen.yml --outfile ${CONTRACTS}/oss.json --type json && \
 rm src/.oss_gen.yml
```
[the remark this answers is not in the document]
glinton: Maybe `tee` could be crudely utilized...
```diff
- swagger-cli bundle src/.oss_gen.yml --type yaml | \
+ swagger-cli bundle src/.oss_gen.yml --type yaml | tee /tmp/oss.log | \
```
Then end the block by `cat`ing the temporary file? not much better, but saves some work. Calling bundle twice isn't very expensive and is ok. Go with whichever you prefer

contracts/cloud.yml
jstirnaman: @danxmoran Does OSS have a token quota or would that be managed by a separate gateway?

>    "429":
      description: The token is temporarily over quota. The Retry-After header describes when to try the write again.
      headers:
        Retry-After:
          description: A non-negative decimal integer indicating the seconds to delay after the response is received.
          schema:
            type: integer
            format: int32

jstirnaman:
@glinton @sanderson @kelseiv @timhallinflux
Please review (or at least glance at the screenshots).
I'll be adding anchor links in the tag descriptions (below the headings) and eventually scripting it - maybe on the Docs side) - to keep it manageable.

sanderson (approved):
LGTM

jstirnaman:
@glinton Can you tell me how to resolve this circleci fail? https://app.circleci.com/pipelines/github/influxdata/openapi/572/workflows/dcbc6f2c-b32e-40a6-8d32-f43fda6a2b42/jobs/2003

scripts/generate.sh
```diff
@@ -1,28 +1,35 @@
...
 CONTRACTS=${CONTRACTS:-contracts}
+INFLUXDB_DOCS_URL=https://docs.influxdata.com/influxdb
 
 # generate oss contract
 sed -e '/#REF_COMMON_PATHS/{r ./src/common/_paths.yml' -e 'd}' src/oss.yml |
 sed -e '/#REF_COMMON_PARAMETERS/{r ./src/common/_parameters.yml' -e 'd}' |
 sed -e '/#REF_COMMON_SCHEMAS/{r ./src/common/_schemas.yml' -e 'd}' > src/.oss_gen.yml && \
-swagger-cli bundle src/.oss_gen.yml --outfile ${CONTRACTS}/oss.yml --type yaml && \
+swagger-cli bundle src/.oss_gen.yml --type yaml | \
+(sed -e "s|{{% INFLUXDB_DOCS_URL %}}|${INFLUXDB_DOCS_URL}/v2.0|g" > ${CONTRACTS}/oss.yml) && \
```
glinton, line 11: `swagger-cli` was creating the `$CONTRACTS` directory, so now you'll have to add a `mkdir -p $CONTRACTS` above line 6.

glinton (changes requested):

scripts/generate.sh
glinton: did you run `make generate-all` as well, or are you manually editing the files in `contracts`?

jstirnaman: Yes, always `make generate-all`. I could never tolerate that much manual editing.

glinton (approved):
:tada: nice!

contracts/cloud-diff.yml
```diff
@@ -2,8 +2,83 @@ openapi: 3.0.0
...
+    description: |
+      Create and manage API tokens. An **authorization** associates a list of permissions to an **organization** and provides a token for API access. To assign a token to a specific user, scope the authorization to the user ID.
+x-tagGroups:
+  - name: Get started
+    tags:
+      - Authentication
+  - name: ' '
+    tags:
+      - Write
+      - Query
```
kelseiv, line 34: prob makes sense to have these under get started with auth? ...the spacing between auth and query/write may look accidental or give pause

PR #195, opened 2021-09-15
feat: restore kv responds with new token after metadata update

lesam:
During a full metadata update, the operator token used to access the database changes (it is part of the metadata). To handle this properly, we need to know the new token value after a successful restore.

See https://github.com/influxdata/influxdb/issues/20749 .


wbaker85 (approved):
LGTM

PR #196, opened 2021-09-20
chore: notebooksd new routes

genehynson:
Adds new routes for notebooksd

### DIFF
--- a/contracts/priv/notebooksd.yml
+++ b/contracts/priv/notebooksd.yml
@@ -290,7 +290,8 @@ paths:
           description: returns result from the pipe's query
           content:
             text/csv:
-              type: string
+              schema:
+                type: string
         '400':
           $ref: '#/components/responses/ServerError'
         '401':
--- a/contracts/priv/notebooksd.yml
+++ b/contracts/priv/notebooksd.yml
@@ -147,6 +147,181 @@ paths:
           $ref: '#/components/responses/ServerError'
         '500':
           $ref: '#/components/responses/ServerError'
+  /notebooks/share:
+    get:
+      summary: get all Share records in org
+      operationId: getAllShares
+      tags:
+        - Share
+      parameters:
+        - name: orgID
+          in: query
+          schema:
+            type: string
+          required: true
+        - name: notebookID
+          in: query
+          schema:
+            type: string
+          required: false
+      responses:
+        '200':
+          description: returns all Share records in org
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Shares'
+        '401':
+          $ref: '#/components/responses/ServerError'
+        '500':
+          $ref: '#/components/responses/ServerError'
+    post:
+      summary: creates a Share
+      operationId: createShare
+      tags:
+        - Share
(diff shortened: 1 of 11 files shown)


contracts/priv/notebooksd.yml
```diff
@@ -147,6 +147,181 @@ paths:
...
+          required: false
+      responses:
+        '200':
+          description: returns all Share records in org
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Shares'
+        '401':
+          $ref: '#/components/responses/ServerError'
```
subirjolly, line 175: Shouldn't be ServerError. It should be Unauthorized.

contracts/priv/notebooksd.yml
```diff
@@ -147,6 +147,181 @@ paths:
...
+        $ref: '#/components/requestBodies/ShareParams'
+      responses:
+        '200':
+          description: Share created
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Share'
+        '400':
+          $ref: '#/components/responses/ServerError'
```
subirjolly, line 193: BadRequest

contracts/priv/notebooksd.yml
```diff
@@ -147,6 +147,181 @@ paths:
...
+        '401':
+          $ref: '#/components/responses/ServerError'
```
subirjolly, line 195: Unauthorized

contracts/priv/notebooksd.yml
```diff
@@ -147,6 +147,181 @@ paths:
...
+        '500':
+          $ref: '#/components/responses/ServerError'
+  '/api/share/{id}/query/{pipeID}':
```
subirjolly, line 271: Just wondering why we chose to go without notebooks in the path. Something like this `/api/notebooks/share/.......`?

contracts/priv/notebooksd.yml
genehynson: I thought so too but it looks like George went through and changed all of them to be "ServerError" in this commit: https://github.com/influxdata/openapi/commit/02f9592832d641eeb00d7ca7bc243693bebcf91d

Check this out: https://github.com/influxdata/openapi/blob/master/src/svc/notebooksd/paths/notebooks_id.yml#L20

contracts/priv/notebooksd.yml
genehynson: well the original plan was to not limit the scope to notebooks. Could be sharing dashboards or something. But now that we are narrowing the scope to notebooks, you're probably right. I'll probably iterate on that in a future PR tho because it will require changes to several places.

contracts/priv/notebooksd.yml
genehynson: Confirmed with Greg that they all should be "ServerError" now. All the error types have the same openapi spec, so he consolidated all the schemas into one.

subirjolly: 😞  Resolving it for now.

subirjolly (approved):

PR #197, opened 2021-09-20
fix: remove conflicting endpoint

genehynson:
Oats was generating `getNotebooksShare` fn for two different endpoints. We don't actively use this one, so I'm removing it for now so it doesn't block integration.


glinton (approved):
gotta love those oats bugs (other generators typically use the `operationId` field to avoid issues like this)

PR #200, opened 2021-09-22
feat(new onboarding) : added new private apis for idempotent onboarding requests

StacieClark:
Added entries for 2 new endponts that looks like this:

PUT /api/v2private/setup/user
Request:src/cloud/schemas/OnboardingWithIDsRequest.yml
{
    "user": { “name” : “username”, “id” : “userId”},
    "org": { “name” : “orgname”, “id” : “orgId”},
    “Config” : {
        "bucket": ”OEM Org Name’s Bucket”,
        "retentionSeconds": 0,
        "limit": {}, // limits struct 
    }
    “returnToken” : true,
}

Response: src/cloud/schemas/OnboardingWithTokenResponse.yml
{
       token: get_in(response, ["auth", "token"]),
       org: {"name": “org_name”, “id”: “org_id”, “links”: “org_links”),
       user: {"name": “user_name”, “id”: “user_id”),

}

PUT /api/v2private/user
Request: src/cloud/schemas/OnboardingUserRequest.yml
{
    user : { “name” : “username”, “id” : “userId”},
    “orgID”: “abcdef1234567890”, 
    “role”: “owner”,             // [owner | member].
}

Response: src/cloud/schemas/OnboardingWithIDsResponse.yml
{
      org: {"name": “org_name”, “id”: “org_id”, “links”: “org_links”),
       user: {"name": “user_name”, “id”: “user_id”),
}


contracts/priv/cloud-priv.yml
```diff
@@ -281,6 +281,123 @@ paths:
...
+          application/json:
+            schema:
+              $ref: '#/components/schemas/OnboardingUserRequest'
+      responses:
+        '201':
+          description: User created
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/OnboardingWithIDsResponse'
```
randycoulman, line 367: Should this be named something like `OnboardingUserResponse` to avoid confusion with the `OnboardingWithIDsRequest` above?

contracts/priv/cloud-priv.yml
```diff
@@ -619,6 +736,66 @@ components:
...
+              type: string
+              description: name of non system bucket
+            retentionPeriodSeconds:
+              type: integer
+              description: retention period of non system bucket
+            limit:
+              $ref: '#/components/schemas/Limit'
+        returnToken:
+          type: bool
+          description: if true will return the auth token in the response
```
randycoulman, line 759: Is `auth token` the right description here? I'm not sure how tokens are named in the platform; this token should be an org-scoped token that allows the OEM customer to manage the new organization, but I'm not sure what the right name for that is.

contracts/priv/cloud-priv.yml
```diff
@@ -619,6 +736,66 @@ components:
...
+        role:
+          type: string
+          description: Role of the User
+          enum:
+            - owner
+            - member
+      required:
+        - user
+        - orgID
+        - Role
```
randycoulman, line 781:
```suggestion
        - role
```

contracts/priv/cloud-priv.yml
```diff
@@ -619,6 +736,66 @@ components:
...
+        org:
+          $ref: '#/components/schemas/IdentityLinks'
+    OnboardingWithTokenResponse:
+      type: object
+      properties:
+        user:
+          $ref: '#/components/schemas/Identity'
+        org:
+          $ref: '#/components/schemas/Identity'
+        auth:
```
randycoulman, line 796: Similar to comment above; should this be named `token` instead? If this is really an `auth token`, then this name is fine.

contracts/priv/cloud-priv.yml
```diff
@@ -619,6 +736,66 @@ components:
...
+        - user
+        - orgID
+        - Role
+    OnboardingWithIDsResponse:
+      type: object
+      properties:
+        user:
+          $ref: '#/components/schemas/Identity'
+        org:
+          $ref: '#/components/schemas/IdentityLinks'
```
randycoulman, line 788:
```suggestion
          $ref: '#/components/schemas/Identity'
```

We don't need links for this response...

contracts/priv/cloud-priv.yml
```diff
@@ -619,6 +736,66 @@ components:
...
+    OnboardingWithTokenResponse:
+      type: object
+      properties:
+        user:
+          $ref: '#/components/schemas/Identity'
+        org:
+          $ref: '#/components/schemas/Identity'
```
randycoulman, line 795:
```suggestion
          $ref: '#/components/schemas/IdentityLinks'
```

...but we do for this one.

contracts/priv/cloud-priv.yml
```diff
@@ -708,6 +885,22 @@ components:
...
         - redirectURL
         - state
+    Identity:
+      type: object
+      properties:
+        id:
+          type: string
+        name:
+          type: string
+    IdentityLinks:
```
randycoulman, line 895: #optional Maybe `IdentityWithLinks` instead for clarity? In the code above, when I saw `IdentityLinks` I was wondering why it was only the links and not the identity as well.

randycoulman (changes requested):
This looks exactly like what we discussed, with one exception below (which response contains the org links).

A few optional suggestions below. The only "Request changes" issue is the org links.

contracts/priv/cloud-priv.yml
StacieClark: That is what is in the specs on the issue

contracts/priv/cloud-priv.yml
StacieClark: What do you want to call it.

contracts/priv/cloud-priv.yml
StacieClark: FYI , anything under contracts is generated from files under /src. So you don't have to repeat the comments if you don't want to :-)

contracts/priv/cloud-priv.yml
StacieClark: We could, although the response is potentially reusable for other APIs. So I name it by what the response contains, not the context of the api from which it is called. if you find it confusing, I can change it

contracts/priv/cloud-priv.yml
StacieClark: What do you want to call it

contracts/priv/cloud-priv.yml
randycoulman: I looked at the API docs, and the endpoints for this are named around the term `authorization`, so this wording is fine. I think I've been working with the JWT/shared session token too much :-).

contracts/priv/cloud-priv.yml
StacieClark: This was an oversight on my part.

contracts/priv/cloud-priv.yml
randycoulman: Let's go with `token`; that's what it's called in the authorizations endpoints.

contracts/priv/cloud-priv.yml
randycoulman: That's a good convention, so I'm fine with it. It just felt a bit odd that an `OnboardingWithIDsResponse` didn't go with an `OnboardingWithIDsRequest`. Ultimately, the names aren't going to be used directly in the Quartz code, so whatever makes sense to you is fine.

contracts/priv/cloud-priv.yml
StacieClark: I change it to token

contracts/priv/cloud-priv.yml
StacieClark: It's with one or the other. An identity with the links, or one without. I've changed it to not have the links. Let me know if this is not what you want because I have to update the Issue with the changed API spec and change my code as well

randycoulman (approved):
Changes LGTM!

src/cloud-priv.yml
```diff
@@ -76,6 +76,32 @@ components:
...
+    OrgLimits:
+      $ref: "./quartz/schemas/OrgLimits.yml"
+    RateLimits:
+      $ref: "./quartz/schemas/RateLimits.yml"
+    RestrictedLimit:
+      $ref: "./quartz/schemas/RestrictedLimit.yml"
+    TaskLimits:
+      $ref: "./quartz/schemas/TaskLimits.yml"
+    Unlimited:
+      $ref: "./quartz/schemas/Unlimited.yml"
```
glinton, line 104: You need to also add something for `./quartz/schemas/Limit.yml` to remove the `2` that is throwing off oats

glinton: It should be able to be named differently if `Limit` is already taken

glinton: made the change since Stacie's out for the weekend (and off next week)

contracts/priv/cloud-priv.yml
```diff
@@ -619,6 +719,66 @@ components:
...
+            retentionPeriodSeconds:
+              type: integer
+              description: retention period of non system bucket
+            limit:
+              $ref: '#/components/schemas/OrgLimits'
+        returnToken:
+          type: boolean
+          description: if true will return the auth token in the response
+      required:
+        - username
```
randycoulman, line 744: Just caught this: should be `user` I think, not `username`

src/cloud/paths/provision.yml
```diff
@@ -0,0 +1,59 @@
...
+          application/json:
+            schema:
+              $ref: "../../common/responses/ServerError.yml"
+        description: Insufficient credentials to create an organization
+      "409":
+        content:
+          application/json:
+            schema:
+              $ref: "../../common/responses/ServerError.yml"
+        description: Conflict. Causes can be user already exists with a different name, org already exists with a different name, org exists with different limits.
```
randycoulman, line 43: Will we also get a 409 when an org name is already in use with a different ID? If not, can/should we?

src/cloud/paths/provision.yml
```diff
@@ -0,0 +1,59 @@
...
+          application/json:
+            schema:
+              $ref: "../schemas/ProvisionResponse.yml"
+        description: The user name and ID, the org name and ID, the token for user if requested
+      "400":
+        content:
+          application/json:
+            schema:
+              $ref: "../../common/responses/ServerError.yml"
+        description: Bad request. Data in request is invalid. Examples include Invalid IDs, invalid names, and invalid limits
```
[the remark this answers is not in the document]
glinton: Not that it's "correct", but I'm fairly certain the rest of the api returns a 400 for these cases; likely why it's documented that way here. Perhaps the plan is to upgrade the entire api at once, vs one service at a time?

src/cloud/paths/provision.yml
StacieClark: Although, maybe that should be 409?

StacieClark: Turns out, due to the conversion of internal codes to Http codes, that I cannot return a 409 at all. Any conflict type will be a 422. I will amend the document

src/cloud-priv.yml
```diff
@@ -62,6 +76,34 @@ components:
...
+    QuartzLimit:
+      $ref: "./quartz/schemas/Limit.yml"
```
randycoulman, line 106: I just realized that these are pointing at the Quartz limit structs. I'm not sure that's correct here. I think this should be IDPE limit structs. They should be roughly the same, and the intention here is that Quartz should do any translation necessary between the two structs. In the past, there hasn't been any translation necessary.

StacieClark: I don't understand why we would have different definitions. This is an object that is used to communicate between quartz and idpe. What other APIs are using these limits?

StacieClark: check it now

randycoulman: Looks good from what I can tell. Let us know when the corresponding change to IDPE goes in (the one that doesn't require us to supply an IDPE ID for the Limit struct).

PR #202, opened 2021-09-24
fix: use empty string for OSS root-path server URL

danxmoran:
Using `/` causes codegen'd clients to generate routes like `//health`. Empty string generates the right thing.

### DIFF
--- a/contracts/cloud.json
+++ b/contracts/cloud.json
@@ -205,7 +205,7 @@
         "summary": "Checks the status of InfluxDB instance and version of InfluxDB.",
         "servers": [
           {
-            "url": "/"
+            "url": ""
           }
         ],
         "tags": [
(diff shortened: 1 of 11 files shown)


DStrand1 (approved):
LGTM, though this seems like a change that may need to be documented somewhere since I can imagine this changes some semantics

jstirnaman:
> Using `/` causes codegen'd clients to generate routes like `//health`. Empty string generates the right thing.

I made the previous change to `'/'` because `''` results in Redoc (the [API reference](https://docs.influxdata.com/influxdb/cloud/api/)) rendering the wrong URL path for the root-level endpoints. It just prepends the default server URL path instead of using the root URL path. I'll look for an alternative way to fix it for generating docs.

PR #209, opened 2021-10-12
fix: Add orgID and bucketID properties to the MeasurementSchema entity

stuartcarnie:
This PR amends the `MeasurementSchema` entity to add the following properties:

* `orgID`,
* `bucketID`

NOTE that the Cloud 2.0 API already returns these values and therefore this PR is ensuring the public API and the spec agree.


jrenee42 (approved):
love all the examples.  

I ran oats locally and it has the two new optional fields; looks great! 
thanks!

PR #210, opened 2021-10-14
fix: managed function has `/api/v2` prefix

bednar:
The managed functions has to have `/api/v2`. Try to:

```
curl -v -X GET 'https://us-west1-1.gcp.cloud2.influxdata.com/api/v2/functions'  --header Authorization <some_token> -H 'accept: application/json'
```

vs

```
curl -v -X GET 'https://us-west1-1.gcp.cloud2.influxdata.com/functions'  --header Authorization <some_token> -H 'accept: application/json'
```

### DIFF
--- a/src/svc-managed-functions.yml
+++ b/src/svc-managed-functions.yml
@@ -1,6 +1,6 @@
 openapi: '3.0.0'
 info:
-  title: Influxdata Managed Functions CRUD API
+  title: InfluxData Managed Functions CRUD API
   version: 0.1.0
 servers:
   - url: '/api/v2'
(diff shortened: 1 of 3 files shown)


src/svc-managed-functions.yml
```diff
@@ -1,9 +1,9 @@
 openapi: '3.0.0'
 info:
-  title: Influxdata Managed Functions CRUD API
+  title: InfluxData Managed Functions CRUD API
   version: 0.1.0
 servers:
-  - url: '/'
+  - url: '/api/v2'
```
glinton, line 6: This is incorrect, but raises an issue with the readme. anything in `/contracts/svc` defines an "internal" api and will likely have a different value for the `servers` key. This needs to be reverted.
```suggestion
  - url: '/'
```

bednar:
This PR is no longer need due the https://github.com/influxdata/openapi/pull/213.

PR #213, opened 2021-10-19
Rename functions scripts

aanthony1243:


glinton (approved):
:writing_hand: rename complete! Just wait to merge until the server has been updated

src/svc/invocable-scripts/schemas/ScriptCreateRequest.yml
```diff
@@ -10,5 +10,5 @@ properties:
     description: script is script to be executed
```
jstirnaman, line 10:
```suggestion
    description: script to be executed
```

src/svc/invocable-scripts/schemas/Scripts.yml
```diff
@@ -1,6 +1,6 @@
 type: object
 properties:
-  runs:
+  scripts:
```
jstirnaman, line 3: @aanthony1243 So `runs` (or log/number/list of invocations) isn't supported right now?

yquansah (approved):
LGTM

contracts/svc/invocable-scripts.yml
```diff
@@ -0,0 +1,317 @@
...
+          enum:
+            - json
+        headers:
+          type: object
+        statusCode:
+          type: string
+          enum:
+            - 200
+            - 500
+            - 404
```
bednar, line 308: Is this necessary? It's not referenced anywhere in the rest of the document.

contracts/svc/invocable-scripts.yml
```diff
@@ -0,0 +1,317 @@
...
+        name:
+          type: string
+        description:
+          type: string
+        script:
+          description: script is script to be executed
+          type: string
+    ScriptHTTPResponseData:
+      description: The data sent to end user when a script is invoked using http. User defined and dynamic
+      type: object
```
bednar, line 281: The current implementation returns response with following HTTP headers:

```
header: Date: Tue, 19 Oct 2021 10:22:19 GMT
header: Content-Type: text/plain; charset=utf-8
header: Transfer-Encoding: chunked
header: Connection: keep-alive
header: Vary: Accept-Encoding
header: trace-id: 78d6cc5dffa0b7d6
header: trace-id: 78d6cc5dffa0b7d6
header: trace-sampled: false
header: trace-sampled: false
header: vary: Accept-Encoding
header: Strict-Transport-Security: max-age=15724800; includeSubDomains
header: X-Influxdb-Build: Cloud
,result,table,_start,_stop,_time,_value,_field,_measurement,language,license,name,owner
...
```

Is `string` type more suitable?

```suggestion
      type: string
```

bednar, line 281: The current implementation returns response with following HTTP headers:

```
header: Date: Tue, 19 Oct 2021 10:22:19 GMT
header: Content-Type: text/plain; charset=utf-8
header: Transfer-Encoding: chunked
header: Connection: keep-alive
header: Vary: Accept-Encoding
header: trace-id: 78d6cc5dffa0b7d6
header: trace-id: 78d6cc5dffa0b7d6
header: trace-sampled: false
header: trace-sampled: false
header: vary: Accept-Encoding
header: Strict-Transport-Security: max-age=15724800; includeSubDomains
header: X-Influxdb-Build: Cloud
,result,table,_start,_stop,_time,_value,_field,_measurement,language,license,name,owner
...
```

Is `string` type more suitable?

```suggestion
      type: string
```

contracts/svc/invocable-scripts.yml
yquansah: `ScriptHTTPResponseData` is used here: https://github.com/influxdata/openapi/blob/371a8ca7e185dfeeafd354a3e7291de695d86d3b/contracts/svc/invocable-scripts.yml#L164

but I do not think we have a use for `ScriptHTTPResponseNoData`

contracts/invocable-scripts.yml
```diff
@@ -0,0 +1,317 @@
...
+      operationId: GetScripts
+      tags:
+        - Scripts
+      summary: List all Scripts
+      parameters:
+        - in: query
+          name: org
+          description: The name of the organization.
+          schema:
+            type: string
```
yquansah, line 19: Actually org name is not supported, only orgID

contracts/svc/invocable-scripts.yml
```diff
@@ -278,7 +278,7 @@ components:
           type: string
     ScriptHTTPResponseData:
       description: The data sent to end user when a script is invoked using http. User defined and dynamic
-      type: object
+      type: string
```
glinton: remember to not manually edit files in `/contracts` but instead update the file in  `/src` and use `make generate-all` to generate the contract

glinton:
> @yquansah @glinton Shouldn't this be included in contracts/cloud as well?

No, `contracts/cloud` and `contracts/oss` are carryover from the single-file-to-describe-the-world days and were never intended to include definitions from new sub services. Here, we opted for granularity for developer satisfaction (extrapolate the "single file" idea out to an api the size of amazon and it's clear to see it isn't a great idea (someone targeting only route53 apis shouldn't be forced into everything else)). For those who demand an all-encompassing file, we still provide them in `contracts/ref/`. Otherwise, the preferred consumption method is to produce client libraries for each subsystem in the api so they can be utilized individually. That doesn't stop us from packaging those clients up into a hierarchical pattern to match programming languages. For example, our golang client libs could be structured something like:
```
/influxdb
  /tasks  \
  /write    | -> currently all coexist inside the cloud/oss definitions, but will one day (hopefully) be split out.
  /query  /
  /scripts
  /annotations
  /any-other-service-specific-client
```

PR #225, opened 2021-11-01
chore: use new swagger aggregator

glinton:
This uses the new version of the swagger aggregator which includes all swagger field keys (including `x-` fields). The downside is the ordering of the resulting swagger file is not as human friendly, but with a file as large as the output, it seems acceptable.


jstirnaman (approved):
Looks good. Thanks!

PR #228, opened 2021-11-02
feat: add 429 to org invites

djcarpe:
The `/invites` endpoint rate limits organization invite post requests and will return `429` ( too many requests ) once the limit is reached.


hansenreid (approved):
LGTM

PR #238, opened 2021-11-19
feat(unity): add account name as property for operator account schema

gspletzer:
For the purpose of multi-account, account names will be a required field for an account. The names need to be added to the Operator UI, so I added the `name` value from the `Account` as a property to the `OperatorAccount` schema in Unity.

### DIFF
--- a/contracts/priv/unity.yml
+++ b/contracts/priv/unity.yml
@@ -1139,6 +1139,9 @@ components:
         id:
           type: number
           description: account id in quartz
+        name:
+          type: string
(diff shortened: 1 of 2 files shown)


randycoulman (approved):
LGTM!

Has this already been added to the `Account` schema? If not, should we add it here as well?

gspletzer:
> LGTM!
> 
> Has this already been added to the `Account` schema? If not, should we add it here as well?

Yes, it has already been added to the Account schema.

PR #242, opened 2021-11-25
fix: merge write specification back together

lesam:


src/common/paths/write.yml
```diff
@@ -114,7 +114,13 @@ post:
     "404":
       $ref: "../../common/responses/ResourceNotFoundError.yml"
     "413":
-      description: Request entity too large. The payload exceeded the 50MB size limit. InfluxDB rejected the batch and did not write any data.
+      description: >
```
lesam, line 117: I think we could do better here with more feature-ful yaml generation, but for now it's quite valuable for us to enforce that API's that are shared by cloud and OSS have the api specification in the same place.

This affects how we do codegen for common tooling (i.e. the CLI). If we allow the configurations to split it is quite easy for them to drift.

src/common/paths/write.yml
```diff
@@ -131,8 +137,16 @@ post:
                     <center>nginx</center>
                   </body>
                 </html>
+        application/json:
```
lesam, line 140: mulitple content types is technically correct here, though unfortunately the openapi-generator codegen tool does not handle it very well. But, it handles it well enough for the CLI.

lesam: In practice, the codegen just has `application/json` take priority, and the schema is `string`, so it supports both well enough for what we need - ugly but it works.

Issue #243, opened 2021-11-29
UI Color Mapping: Add a property to XYViewProperties to enable color mapping


hoorayimhelping:
Swagger will need to be updated to handle the new color mapping property (this ticket leaves the name intentionally ambiguous to allow for the implementer to name it appropriately) before IDPE and UI can make use of it.

PR #248, opened 2021-12-15
feat: add colorMapping property type to the swagger definition

ChitlangeSahas:
Closes:  https://github.com/influxdata/openapi/issues/243
Depends on: https://github.com/influxdata/idpe/pull/12366

This PR adds the colorMapping property to the swagger definition. I ran `make generate` to update the other files needed.


src/common/schemas/XYViewProperties.yml
```diff
@@ -24,6 +24,9 @@
       type: array
       items:
         $ref: "./DashboardColor.yml"
+    colorMapping:
+      description: An object that contains information about the color mapping
+      $ref: "./ColorMapping.yml"
```
ChitlangeSahas, line 29: These changes are mine.

src/common/schemas/ColorMapping.yml
```diff
@@ -0,0 +1,7 @@
+  type: object
+  description: maps series id to the color index that is used to index into the colors array
+  additionalProperties:
+    type: int
+  example:
+    series_id_1: 0
+    series_id_2: 1
```
ChitlangeSahas, line 7: And these. Rest are generated.

contracts/common.yml
```diff
@@ -9863,6 +9866,14 @@ components:
       enum:
         - implicit
         - explicit
+    ColorMapping:
+      type: object
+      description: maps series id to the color index that is used to index into the colors array
```
hoorayimhelping, line 9871: I think this is a bit of an obvious comment. I think a better description would explain what a color mapping is so that someone reading this definition who wasn't familiar with color mapping would understand what it's for.

>A color mapping is an object that maps time series data to a UI color scheme to allow the UI to render graphs consistent colors across reloads.

contracts/common.yml
```diff
@@ -9863,6 +9866,14 @@ components:
...
+      additionalProperties:
+        type: int
+      example:
+        series_id_1: 0
+        series_id_2: 1
```
hoorayimhelping, line 9876: this is a decent example, but it might be more illustrative if the series ids were more like how they'll be in real life using a common example that many people in the company would be familar with:

```
measurement_birdmigration_europe: 0
```

that isn't a very good example, but I'm sure you can find a more common example using sample data or something like that?

hoorayimhelping (commented):
this is solid and will work just fine, but i gave a couple suggestions to make the documentation a little bit richer. let me know what you think

hoorayimhelping (approved):
thanks for updating those examples! solid work 👍

PR #249, opened 2021-12-16
fix: v1compat and auth schemes

jstirnaman:
* Fixes auth (security) schemes in cloud, oss, and v1compat.
* Adds missing `db=` and `rp=` query params in v1compat `/query`.
* Adds tagging for navigation in v1compat API docs.

### DIFF
--- a/contracts/cloud-diff.yml
+++ b/contracts/cloud-diff.yml
@@ -9,10 +9,13 @@ servers:
 tags:
   - name: Authentication
     description: |
-      Use one of the following schemes to authenticate to the InfluxDB API:
-      - [Token authentication](#section/Authentication/TokenAuthentication)
-      - [Basic authentication](#section/Authentication/BasicAuthentication)
-      - [Querystring authentication](#section/Authentication/QuerystringAuthentication)
+      The InfluxDB `/api/v2` API requires authentication for all requests.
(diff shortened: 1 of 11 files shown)


jstirnaman:
Cloud and OSS `/api/v2/me/password`, `/api/v2/users/{userID}/password`, and `/api/v2/signin` paths reference securityScheme `BasicAuthentication`, meaning those APIs support Basic auth. Is that correct?

glinton (approved):
Looks fine, don't forget to run `make generate-all` before committing to clear up that test :slightly_smiling_face:

wbaker85:
For OSS, `/api/v2/signin` does use `BasicAuthentication` to create a session from the provided user name/password.

`/api/v2/me/password` does not work at all. It's just implemented in a way such that it could never work correctly. I think we should actually eliminate this route from the API spec as `/api/v2/users/{userID}/password` is the correct endpoint to use for this, and is what the CLI uses. I don't think these endpoints are used on cloud for user management currently unless @glinton knows otherwise?

`/api/v2/users/{userID}/password` actually uses token authentication like everything else. The user ID comes from the URL and the request body contains a `password` field that the user's password will be updated to if the token has the correct permissions.

jstirnaman:
@wbaker85 Thanks for looking into this and explaining. I'll create some issues and fix what I can.

PR #254, opened 2021-12-21
feat: enable remotes/replications resources for OSS

wbaker85:
Uncomments the appropriate sections since remotes & replications are now on by default in OSS.


DStrand1 (approved):
LGTM

PR #255, opened 2021-12-21
feat: config endpoint for oss

wbaker85:
Part of https://github.com/influxdata/influxdb/issues/22917


davidby-influx (approved):
LGTM, but I would recommend a second reviewer because of my inexperience with this technology...

DStrand1 (approved):
LGTM! But is this something that is incompatible/undesirable for Cloud? Wondering if it should be for both

wbaker85:
> LGTM! But is this something that is incompatible/undesirable for Cloud? Wondering if it should be for both

It's incompatible with cloud - there's no concept of a user-defined configuration file there.

PR #259, opened 2022-01-13
chore: added example property to fluxdocs

blegesse-git:
closes influxdata/ui#3584 

Added a new property to fluxdocs

### DIFF
--- a/contracts/priv/fluxdocsd.yml
+++ b/contracts/priv/fluxdocsd.yml
@@ -147,6 +147,7 @@ components:
         - kind
         - headline
         - fluxType
+        - example
       properties:
         path:
           type: string
@@ -161,6 +162,8 @@ components:
         fluxType:
           type: string
         description:
+          type: string
+        example: 
           type: string
         fluxParameters:
           type: array
--- a/contracts/priv/fluxdocsd.yml
+++ b/contracts/priv/fluxdocsd.yml
@@ -163,7 +163,7 @@ components:
           type: string
         description:
           type: string
-        example: 
+        example:
           type: string
         fluxParameters:
           type: array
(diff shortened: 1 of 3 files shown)


contracts/priv/fluxdocsd.yml
```diff
@@ -147,6 +147,7 @@ components:
         - kind
         - headline
         - fluxType
+        - example
```
TCL735, line 150: I don't think we need this on the backend. Let's make it optional rather than required by removing it from the required list.

contracts/svc/fluxdocsd.yml
```diff
@@ -147,6 +147,7 @@ components:
         - kind
         - headline
         - fluxType
+        - example
```
TCL735, line 150: I don't think we need this on the backend. Let's make it optional rather than required by removing it from the required list.

src/svc/fluxdocsd/schemas/fluxdocs.yml
```diff
@@ -5,6 +5,7 @@ required:
   - kind
   - headline
   - fluxType
+  - example
```
TCL735, line 8: I don't think we need this on the backend. Let's make it optional rather than required by removing it from the required list.

TCL735 (changes requested):

src/svc/fluxdocsd/schemas/fluxdocs.yml
blegesse-git: is there going to be a case where a flux function will not have an example value? it seems like they all should have this property..

TCL735:
Closing this. Will do https://github.com/influxdata/ui/issues/3638 instead.

PR #261, opened 2022-01-24
fix: update nifi

helenosheaa:
Closes https://github.com/influxdata/data-acquisition/issues/99

Updates for our latest api definition for nifid


src/svc/nifid/schemas/Subscription.yml
```diff
@@ -10,25 +12,73 @@ properties:
...
+        type: string
+  stringFields:
+    type: array
+    items:
+      type: object
+      properties:
+        pattern:
+          type: string
+        name:
+          type: string
```
helenosheaa, line 61: based on how we've defined our yml, jsonspec is it's own schema whereas the string/field & tag object is being defined inline, I suppose it's not an issue but just checking?

helenosheaa: fixed in commit 292ddc7

genehynson (approved):

PR #263, opened 2022-01-31
fix: schema violations and grammar

jstirnaman:
Fix invalid `type: timestamp` by replacing with `type: integer, format: unix timestamp`.
Remove "required" error response attributes that aren't in all errors.
Add missing OperationIDs.
Fix grammar in descriptions.

### DIFF
--- a/src/common/schemas/Error.yml
+++ b/src/common/schemas/Error.yml
@@ -3,14 +3,15 @@
       $ref: "./ErrorCode.yml"
     message:
       readOnly: true
-      description: message is a human-readable message.
+      description: Human-readable message.
       type: string
     op:
       readOnly: true
-      description: op describes the logical code operation during error. Useful for debugging.
+      description: Describes the logical code operation when the error occurred. Useful for debugging.
       type: string
     err:
       readOnly: true
-      description: err is a stack of errors that occurred during processing of the request. Useful for debugging.
+      description: Stack of errors that occurred during processing of the request. Useful for debugging.
       type: string
-  required: [code, message]
+  required:
+    - code
--- a/src/common/schemas/LineProtocolError.yml
+++ b/src/common/schemas/LineProtocolError.yml
@@ -3,19 +3,20 @@
       $ref: "./LineProtocolErrorCode.yml"
     message:
       readOnly: true
-      description: Message is a human-readable message.
+      description: Human-readable message.
       type: string
     op:
       readOnly: true
-      description: Op describes the logical code operation during error. Useful for debugging.
+      description: Describes the logical code operation when the error occurred. Useful for debugging.
       type: string
     err:
       readOnly: true
-      description: Err is a stack of errors that occurred during processing of the request. Useful for debugging.
+      description: Stack of errors that occurred during processing of the request. Useful for debugging.
       type: string
     line:
       readOnly: true
-      description: First line within sent body containing malformed data
+      description: First line in the request body that contains malformed data.
       type: integer
       format: int32
-  required: [code, message, op, err]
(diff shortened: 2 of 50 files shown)


glinton (approved):

src/common/schemas/Error.yml
```diff
@@ -3,14 +3,15 @@
...
+      description: Describes the logical code operation when the error occurred. Useful for debugging.
       type: string
     err:
       readOnly: true
-      description: err is a stack of errors that occurred during processing of the request. Useful for debugging.
+      description: Stack of errors that occurred during processing of the request. Useful for debugging.
       type: string
-  required: [code, message]
+  required:
+    - code
```
jstirnaman, line 17: @glinton @lesam Can you verify this change is OK? Just calling it out so I don't break a client.

src/common/schemas/LineProtocolError.yml
```diff
@@ -3,19 +3,20 @@
...
       type: string
     line:
       readOnly: true
-      description: First line within sent body containing malformed data
+      description: First line in the request body that contains malformed data.
       type: integer
       format: int32
-  required: [code, message, op, err]
+  required:
+    - code
```
jstirnaman, line 22: @glinton @lesam Can you verify this change is OK? Just calling it out so I don't break a client.

src/common/schemas/Error.yml
glinton: Thanks for the callout here! I'd be more concerned about the ui client generated from oats, as my guess is that is what drove those to be required fields.

src/common/schemas/Task.yml
```diff
@@ -4,44 +4,61 @@
...
+        Value is a [duration literal](https://docs.influxdata.com/flux/v0.x/spec/lexical-elements/#duration-literals)).
       type: string
+      format: duration
     cron:
-      description: A task repetition schedule in the form '* * * * * *'; parsed from Flux.
+      description: >-
+        [Cron expression](https://en.wikipedia.org/wiki/Cron#Overview) that defines the schedule on which the task runs.
+        Cron scheduling is based on system time.
+
+        Value is a [Cron expression](https://en.wikipedia.org/wiki/Cron#Overview).
```
lesam, line 47: I get these errors from our codegen for the cli:

```
Error parsing /openapi/src/common/schemas/Task.yml: bad indentation of a mapping entry (39:37)

 36 |  ... type: string
 37 |  ... format: duration
 38 |  ... on:
 39 |  ... description: [Cron expression](https://en.wikipedia.org/wiki/ ...
-----------------------------------------^
 40 |  ... type: string
 41 |  ... fset:
"/openapi/contracts/oss.yml" is not a valid JSON Schema
```

lesam (changes requested):
I get a bunch of new errors when running `make generate-all` on this change as the [./README.md](./README.md) suggests. Seems like we have some missing CI checks...

lesam:
@DStrand1 FYI

jstirnaman:
> 

@glinton @lesam I didn't get any errors from `make generate-all` on my machine.

glinton:
unless something changed, ci should complain if `make generate-all` wasn't run locally before changes were committed. i'm curious if there is some encoding difference on your machine to get those errors. i can't reproduce them either

lesam:
That's really weird - I somehow had an old version of the branch. With a fresh pull everything works. This will require some changes to the CLI FYI @DStrand1 - since 'Message' is no longer required it becomes a `*string` instead of a `string` in the generated code. But I don't think that has to block the PR.

PR #269, opened 2022-02-08
chore: remove token from response/request bodies.

subirjolly:
Removed token from `POST` and removed `PUT` route from notebooksd. Need to remove token from openapi request/response bodies.


genehynson (approved):

genehynson:
I think you need to run `make generate-all`

genehynson (approved):

PR #270, opened 2022-02-08
✨ Make separate Operator Org Limits

gspletzer:
Part of Quartz #5810

To add `queryTime` to Operator view, new schemas were needed for `OperatorOrgLimits` and `OperatorRateLimits`. Additionally the schema for `Limit` in the `Cloud` context to include `queryTime` as it is used in the `LimitsReducer` to update limits in the Operator view. All files that reference these schemas were updated as well.

### DIFF
--- a/contracts/cloud-diff.yml
+++ b/contracts/cloud-diff.yml
@@ -3329,6 +3329,9 @@ components:
         rate:
           type: object
           properties:
+            queryTime:
+              type: integer
+              description: Query Time in nanoseconds
             readKBs:
               type: integer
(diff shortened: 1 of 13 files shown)


randycoulman (approved):
It's not clear to me where all of these Limit structs are used in the UI and whether this would break any of them.

Is there any difference between the Cloud Limit definition and the Operator Limit definition? If not, should those be shared instead?

Other than that, it does make sense to separate the "private" operator limits shape from the OEM API's definition, so that part is definitely fine. And I'm not sure we want to "cross the streams" between Quartz API specs and IDPE's.

PR #271, opened 2022-02-09
chore(nifid): remove token, add tokenID

genehynson:
Removes `token` from subscriptionParams, adds `tokenID` to subscription response


helenosheaa (approved):
lgtm

PR #273, opened 2022-02-16
chore: update `api/v2/template` endpoint for new 422 response

DStrand1:
Updates the OpenAPI spec to conform to the new templates apply endpoint responses, as shown here: https://github.com/influxdata/influxdb/pull/23121, and a similar change made to Cloud


src/common/schemas/TemplateSummaryError.yml
```diff
@@ -0,0 +1,619 @@
+  type: object
```
lesam, line 1: Can we use `allOf` to make this more DRY ? (Don't-Repeat-Yourself)

lesam (approved):

PR #279, opened 2022-02-23
feat: add max age field to replications create

DStrand1:
Part of https://github.com/influxdata/influxdb/issues/23109

The default chosen here, being `168 * time.Hour` (or 1 week) is adapted from Hinted Handoff, where this feature already exists

### DIFF
--- a/contracts/oss.yml
+++ b/contracts/oss.yml
@@ -13237,6 +13237,11 @@ components:
         dropNonRetryableData:
           type: boolean
           default: false
+        maxAge:
+          type: integer
+          format: int64
+          minimum: 0
+          default: 604800000000000
       required:
         - name
         - orgID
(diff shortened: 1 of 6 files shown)


glinton (approved):

contracts/oss.yml
```diff
@@ -13237,13 +13237,19 @@ components:
         dropNonRetryableData:
           type: boolean
           default: false
+        maxAge:
```
lesam, line 13240: What unit is this in - should we have a comment? And do we usually use seconds or nanoseconds?

lesam: See for example:

```
    retentionPeriodSeconds:
      type: integer
      format: int64
    retentionPeriodHrs:
      type: integer
      deprecated: true
      description: >
        Retention period *in nanoseconds* for the new bucket. This key's name has been misleading since OSS 2.0 GA,
        please transition to use `retentionPeriodSeconds`
```

lesam: Considering our retention period is in seconds in might make sense to call this `maxAgeSeconds`

lesam (approved):

PR #291, opened 2022-03-18
feat(13058 add patch method on limits): add PATCH method to private limits path

StacieClark:
Add PATCH method to private limits path.
Note that the request body is NOT a Limit object like it is for PUT. It's a LimitPatch which is json made up any part of the Limit struct. Only data in the patch is used to overwrite or add data to the limits. No defaults are used.


src/cloud/schemas/LimitPatch.yml
```diff
@@ -0,0 +1,75 @@
...
+        description: comma separated list of notification rules
+        example: 'http,pagerduty'
+  notificationEndpoint:
+    type: object
+    properties:
+      blockedNotificationEndpoints:
+        type: string
+        description: comma separated list of notification endpoints
+        example: 'http,pagerduty'
+    required: ['blockNotificationEndpoints']
```
wbaker85, line 75: I noticed that the `Limit.yml` schema has a [`features`](https://github.com/influxdata/openapi/blob/6ea7df4daa5735a063be3db60d0165b34b26c096/src/cloud/schemas/Limit.yml#L88) property as well. Is that intentionally not included here?

glinton (approved):

src/cloud/schemas/LimitPatch.yml
StacieClark: Intentionally not included. There is no features field on idpe.Limit. The api/v2private endpoint doesn't look for that field in the request either.

wbaker85 (approved):

PR #298, opened 2022-04-04
feat: add query parameters to tasks contract that fulfill the pagination, filtering, and sorting epic

yquansah:
There has been an effort to support multiple changes on the server side for support of pagination, filtering, and sorting of tasks on the server side, to get away from doing that on the client side.

This PR adds some documentation for the parameters that were added for query on the API.

### DIFF
--- a/src/cloud/paths/tasks.yml
+++ b/src/cloud/paths/tasks.yml
@@ -48,6 +48,7 @@ get:
       description: The non-zero number of tasks to return
     - in: query
       name: offset
+      required: false
       schema:
         type: integer
         minimum: 0
@@ -56,6 +57,7 @@ get:
     - in: query
       name: sortBy
       description: Field that records should be sorted by
+      required: false
       schema:
         type: string
         enum:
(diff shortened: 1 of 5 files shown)


glinton (approved):

src/cloud/paths/tasks.yml
```diff
@@ -46,6 +46,20 @@ get:
...
+        minimum: 0
+        default: 0
+      description: The number of records to skip
+    - in: query
+      name: sortBy
+      description: Field that records should be sorted by
+      schema:
+        type: string
+        enum:
+          - name
```
hoorayimhelping, line 62: can we mark these as not required so that the UI not sending them over won't change current behavior?

hoorayimhelping (commented):
just one comment/question about these parameters being required

src/cloud/paths/tasks.yml
yquansah: @hoorayimhelping Yes, I will do that!

gavincabbage:
@yquansah Are we good to merge this?

yquansah:
@gavincabbage we are good to merge this. This PR slipped through the cracks.

PR #302, opened 2022-04-07
feat: add yml contracts for new service uiproxyd

Palakp41:


src/svc-uiproxyd.yml
```diff
@@ -0,0 +1,18 @@
+openapi: "3.0.0"
+info:
+  title: UI Proxy service
+  version: 0.1.0
+servers:
+  - url: /api/v1
+paths:
+  /mapToken:
+    $ref: "./svc/mapsd/paths/mapToken.yml"
```
glinton, line 9: did you mean for this (and line 13) to point to the new files in this pr, rather than `mapsd/`?

glinton (approved):

src/svc-uiproxyd.yml
Palakp41: yes, i just caught it and updated it.

PR #303, opened 2022-04-11
Add Quartz `GET /identity` endpoint spec

eatondustin1:
This adds the Quartz `GET /api/v2private/identity` spec.

### DIFF
--- a/src/unity/schemas/IdentityAccount.yml
+++ b/src/unity/schemas/IdentityAccount.yml
@@ -1,0 +1,18 @@
+properties:
+  id:
+    type: number
+    description: account id in quartz
+  name:
+    type: string
+    description: name of the account
+  type:
+    $ref: './AccountType.yml'
+  paygCreditStartDate:
+    type: string
+    nullable: true
+    description: a timestamp indicates if an account is part of the payg credit experiment, nil if not
+required:
+  - id
+  - name
+  - type
+  - paygCreditStartDate
--- a/src/unity/schemas/IdentityAccount.yml
+++ b/src/unity/schemas/IdentityAccount.yml
@@ -10,9 +10,8 @@ properties:
   paygCreditStartDate:
     type: string
     nullable: true
-    description: a timestamp indicates if an account is part of the payg credit experiment, nil if not
+    description: an ISO8601 timestamp indicating if an account is part of the payg credit experiment, nil if not
 required:
   - id
   - name
   - type
-  - paygCreditStartDate
--- a/src/unity/schemas/IdentityOrganization.yml
+++ b/src/unity/schemas/IdentityOrganization.yml
@@ -1,7 +1,4 @@
 properties:
-  id:
-    type: number
-    description: quartz ID of the org
   idpeId:
     type: string
     description: idpe ID of the org
--- a/src/unity/schemas/IdentityOrganization.yml
+++ b/src/unity/schemas/IdentityOrganization.yml
@@ -1,0 +1,18 @@
+properties:
+  id:
+    type: number
+    description: quartz ID of the org
+  idpeId:
+    type: string
+    description: idpe ID of the org
+  name:
+    type: string
+    description: name of the org
+  clusterHost:
+    type: string
+    description: the url of the current cluster
+required:
+  - id
+  - idpeId
+  - name
+  - clusterHost
--- a/src/unity/schemas/IdentityOrganization.yml
+++ b/src/unity/schemas/IdentityOrganization.yml
@@ -1,5 +1,5 @@
 properties:
-  idpeId:
+  id:
     type: string
     description: idpe ID of the org
   name:
@@ -10,6 +10,5 @@ properties:
     description: the url of the current cluster
 required:
   - id
-  - idpeId
   - name
   - clusterHost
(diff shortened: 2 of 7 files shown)


src/unity/schemas/IdentityAccount.yml
```diff
@@ -0,0 +1,18 @@
...
+    description: account id in quartz
+  name:
+    type: string
+    description: name of the account
+  type:
+    $ref: './AccountType.yml'
+  paygCreditStartDate:
+    type: string
+    nullable: true
+    description: a timestamp indicates if an account is part of the payg credit experiment, nil if not
```
randycoulman, line 13: Maybe mention the date format here? Something like ISO8601 or whatever it is.

src/unity/schemas/IdentityAccount.yml
```diff
@@ -0,0 +1,18 @@
...
+required:
+  - id
+  - name
+  - type
+  - paygCreditStartDate
```
randycoulman, line 18: If `nil` is possible for this field, then maybe we don't need to mark it as `required` so it can be safely omitted?

src/unity/schemas/IdentityOrganization.yml
```diff
@@ -0,0 +1,18 @@
+properties:
+  id:
+    type: number
+    description: quartz ID of the org
```
randycoulman, line 4: Do we need `id` here? I don't think we want to confuse the UI with two IDs; I think we can safely isolate that bit of complexity to Quartz.

src/unity/schemas/IdentityUser.yml
```diff
@@ -0,0 +1,37 @@
...
+    description: user's id in idpe
+  email:
+    type: string
+    description: user's email
+  firstName:
+    type: string
+    description: user's first name
+  lastName:
+    type: string
+    description: user's last name
```
randycoulman, line 16: What are we going to return for users that don't have first/last name? `null` or empty string? If `null`, then these need to be nullable. If empty string, then perhaps we should mention that here so that the UI team knows that they need to address that case.

randycoulman (approved):
Overall, this looks like what we discussed. A few suggestions below. All optional, but I suggest strong considering not exposing Quartz IDs except where necessary (accounts).

src/unity/schemas/IdentityUser.yml
abshierjoel: We went ahead and updated these to be nullable. We did reach out to the UI team to see how they would like to handle this and are pending a response.

src/unity/schemas/IdentityOrganization.yml
abshierjoel: We eliminated the `id` field here and in `IdentityUser.yml`. Just for some context, the `User.yml` has and `id` and an `idpeId`, so we kept the naming the same as that for consistency. In the future, it may be better to rename `idpeId` to just `id` and remove the Quartz ID from all instances of User & Org.

randycoulman: Hmmm, that's unfortunate. I wonder if the UI uses that ID for anything.

While I do like the consistency you've gone for, I'm wondering if we should make this payload be what we want going forward and to a separate issue/PR to make the other usage consistent with this one.  Might be one less breaking change to the API that way when we fix this up later. Up to you, though.

src/unity/schemas/IdentityOrganization.yml
```diff
@@ -0,0 +1,15 @@
...
+    type: string
+    description: idpe ID of the org
+  name:
+    type: string
+    description: name of the org
+  clusterHost:
+    type: string
+    description: the url of the current cluster
+required:
+  - id
```
randycoulman, line 12: This line should no longer be here, since there isn't an `id` field any more. Unless we rename `idpeId` to `id`, then we'd want to delete `idpeId` from the line below.

src/unity/schemas/IdentityUser.yml
```diff
@@ -0,0 +1,34 @@
+properties:
+  idpeId:
```
randycoulman, line 2: I think we should consider renaming this `id`, because that's the only ID there is from the client's perspective.

src/unity/schemas/IdentityUser.yml
```diff
@@ -0,0 +1,34 @@
...
+    description: role permission whether this operator is "read-only" | "read-write"
+  accountCount:
+    type: integer
+    description: number of accounts the user belongs to
+  orgCount:
+    type: integer
+    description: number of orgs the user belongs to within their active account
+required:
+  - id
+  - idpeId
```
randycoulman, line 31: This field list doesn't match with what's above; we need to remove either `id` or `idpeId` from here depending on how we decide to name IDs.

randycoulman (changes requested):
There are still a few more tweaks required here.

src/unity/schemas/IdentityUser.yml
abshierjoel: I went ahead and renamed both of these to just `id` (`IdentityUser` and `IdentityOrganization`)

randycoulman (approved):
Changes LGTM!

gspletzer:
alchemy please merge

PR #304, opened 2022-04-14
Multi-org Organization Spec changes

hansenreid:
Renames old Organization that was being used by operator pages to OperatorOrganization.
Creates a new Organization(s) schema that will be used in multi-org.
Adds paths to get one, get a list and patch one Organization.

part of: https://github.com/influxdata/quartz/issues/6138
closes: https://github.com/influxdata/quartz/issues/6139


src/unity.yml
```diff
@@ -154,12 +158,16 @@ components:
       $ref: './unity/schemas/Me.yml'
     OrgUsers:
       $ref: './unity/schemas/OrgUsers.yml'
+    OperatorOrganization:
+      $ref: './unity/schemas/OperatorOrganization.yml'
+    OperatorOrganizations:
+      $ref: './unity/schemas/OperatorOrganizations.yml'
+    RelatedAccount:
```
randycoulman, line 165: Should this stay where it was for 🔤  order?

src/unity/paths/operator_accounts_accountId.yml
```diff
@@ -24,7 +24,7 @@ get:
                   organizations:
                     type: array
                     description: organizations in the account
-                    $ref: '../schemas/Organizations.yml'
+                    $ref: '../schemas/OperatorOrganizations.yml'
```
randycoulman, line 27: This and the other changes like it will be a breaking change in the UI, so we'll need to update the UI immediately after this gets merged.

Would it be safer to split this rename out to a separate PR, get it merged and integrated into the UI, then add the new schemas that have these same names?

src/unity/schemas/Organization.yml
```diff
@@ -1,23 +1,37 @@
 type: object
 properties:
-  idpeId:
-    type: string
-    description: idpe ID of the org
   id:
-    type: number
-    description: quartz ID of the org
+    type: string
+    description: ID of the org
```
randycoulman, line 5:
```suggestion
    description: IDPE ID of the org
```

#nit

src/unity/schemas/Organization.yml
```diff
@@ -1,23 +1,37 @@
...
+  description:
+    type: string
+    description: the description of the organization
+  creationDate:
+    type: string
+    format: date-time
+    description: the date the organization was created
+  regionCode:
+    type: string
+    description: the code associated with the region
```
randycoulman, line 18:
```suggestion
    description: the canonical code for the region where the organization is hosted
```

src/unity/schemas/Organization.yml
```diff
@@ -1,23 +1,37 @@
...
+  regionName:
     type: string
-    description: region of the org
+    description: the name of the region
```
randycoulman, line 21:
```suggestion
    description: the name of the region where the organization is hosted
```

src/unity/schemas/Organization.yml
```diff
@@ -1,23 +1,37 @@
...
-    $ref: './RelatedAccount.yml'
-required: [account, id, idpeId, region, provider, date]
+    description: the url of the current cluster
+  isActive:
+    type: boolean
+    description: is this the currently active organization in the session?
+  isDefault:
+    type: boolean
+    description: is this the user's default organization in this account?
+required: [id, clusterHost, isRegionBeta, regionCode, regionName, isActive, isDefault]
```
randycoulman, line 37: Are we going to return `isActive` and `isDefault` from the `GET /orgs/:org_id` endpoint? If not, then should these fields be marked as "not required"?

randycoulman (approved):
LGTM! A few questions/suggestions below.

Make sure to time the merging of this to when we can quickly deal with the resulting changed typenames in the generated UI code.

src/unity.yml
eatondustin1, line 165: Do you think it's worth renaming `RelatedAccount` to `OperatorRelatedAccount` as well just so it's clear what it belongs to?

src/unity/schemas/Organization.yml
```diff
@@ -1,23 +1,37 @@
...
   id:
-    type: number
-    description: quartz ID of the org
+    type: string
+    description: ID of the org
   name:
     type: string
     description: name of the org
-  region:
+  description:
```
eatondustin1, line 9: We haven't added the description to the Quartz DB yet because we weren't confident it was even used, so we might not want to add this just yet. Especially since it will require a trip to IDPE until we move it to Quartz.

src/unity/schemas/Organization.yml
```diff
@@ -1,23 +1,37 @@
...
+    type: boolean
+    description: is the region of the organization in beta
+  clusterHost:
     type: string
-    description: date org was created
-  account:
-    $ref: './RelatedAccount.yml'
-required: [account, id, idpeId, region, provider, date]
+    description: the url of the current cluster
+  isActive:
```
eatondustin1, line 31: I think we're going to want a separate schema (something like `OrganizationSummary`) for the `GET /orgs` list endpoint that just returns the `id`, `name`, `isActive`, and `isDefault`. Then we can remove `isActive` and `isDefault` from this model. Otherwise the list endpoint will be returning way more information than needed, especially as we start moving/adding more org data to Quartz.

eatondustin1 (changes requested):
Looks good, I just have question, one suggestion, and one thing I think we'll need to change below.

src/unity/schemas/Organization.yml
eatondustin1: I personally think the `GET /orgs` and `GET /orgs/:idpe_id` should use different schemas, otherwise we can't properly enforce what should and shouldn't be required. Or we end up returning extra information on the list endpoint.

src/unity/schemas/Organization.yml
randycoulman: @eatondustin1 I added to all of the stories; I think we'll take ownership of the field since it exists in IDPE. I added a story to add the field to the table as well.

src/unity.yml
gspletzer: There didn't appear to be 🔤 order in the file, so that is why we didn't move it.

src/unity/schemas/Organization.yml
hansenreid: What other org data is going to be added to quartz? I don't think it is worth adding a new schema to get rid of two fields. If we know for sure there are a lot of fields that will be added, then I agree with you. However, if we don't know what those fields are / will be, I would prefer to wait to split out the schemas. I can also see a situation where we get the list of orgs for a user and would like to know which ones are active / are the default.

src/unity.yml
hansenreid: In theory, `RelatedAccount` can be used for more than just operator, so I am not sure it needs to be renamed

src/unity/schemas/Organization.yml
hansenreid: Updated in https://github.com/influxdata/openapi/pull/304/commits/0346df21a2f03e7163b1469b1c957660c3903cfc

src/unity/schemas/Organization.yml
hansenreid: Updated in https://github.com/influxdata/openapi/pull/304/commits/0346df21a2f03e7163b1469b1c957660c3903cfc

src/unity/schemas/Organization.yml
hansenreid: Updated in https://github.com/influxdata/openapi/pull/304/commits/0346df21a2f03e7163b1469b1c957660c3903cfc

src/unity/schemas/Organization.yml
randycoulman: `isActive` and `isDefault` take extra work to compute and inject into the structs that we return. I know we debated using a payload like this for the org list endpoint:

```json
{
  activeOrgId: 1234,
  defaultOrgId: 5678,
  orgs: [...]
}
```

but we instead chose to put the `isActive` and `isDefault` fields into the org structs instead, as we think it's easier to consume on the UI side.

The UI doesn't likely need those fields when querying a single org.

Similarly, the cluster information (provider/region*) takes extra work to compute, so we probably only want to include that in the single-org endpoint and not in the list.

Can we just do GraphQL now?

hansenreid: I wish... In that case we will go ahead and split out the schemas.

hansenreid:
Split out `Organizations` to `OrganizationSummaries` in https://github.com/influxdata/openapi/pull/304/commits/9376699404af366f6f0b286f9d34caf4ad819ed3

src/unity.yml
eatondustin1: I feel like we could say the same thing for the schemas that are already prefixed with `Operator`, but that all of them will very likely not be used outside of the operator API, so that's why I suggested it, but I'm fine with not renaming it too.

src/unity/paths/orgs_orgId.yml
```diff
@@ -0,0 +1,60 @@
...
+  tags:
+    - Organizations
+  summary: Get an organization
+  parameters:
+    - in: path
+      name: orgId
+      schema:
+        type: string
+      required: true
+      description: The ID of the organization to get.
```
eatondustin1, line 12:
```suggestion
      description: The IDPE ID of the organization to get.
```

src/unity/paths/orgs_orgId.yml
```diff
@@ -0,0 +1,60 @@
...
+      application/json:
+        schema:
+          $ref: "../schemas/OrganizationUpdate.yml"
+  parameters:
+    - in: path
+      name: orgId
+      schema:
+        type: string
+      required: true
+      description: The ID of the organization to update.
```
eatondustin1, line 44:
```suggestion
      description: The IDPE ID of the organization to update.
```

src/unity.yml
```diff
@@ -154,12 +158,16 @@ components:
...
-      $ref: './unity/schemas/Organizations.yml'
+    OperatorOrganization:
+      $ref: './unity/schemas/OperatorOrganization.yml'
+    OperatorOrganizations:
+      $ref: './unity/schemas/OperatorOrganizations.yml'
     RelatedAccount:
       $ref: './unity/schemas/RelatedAccount.yml'
+    Organization:
+      $ref: './unity/schemas/Organization.yml'
+    OrganizationSummaries:
```
eatondustin1, line 169: This is just a question for my benefit, why do we include `OrganizationSummaries` (plural) here but not `OrganizationSummary` (singular)? Do we only need to include the top-level schemas that are referenced by the paths?

eatondustin1 (changes requested):
Looks good, just a couple of places that still use `ID` instead of `IDPE ID` and one question I had.

eatondustin1 (approved):
🚀

PR #305, opened 2022-04-19
✨ Add `users` to create OEM Org endpoint

abshierjoel:
This adds an optional list of user's emails to the OEM Post Organization request body.


djcarpe (approved):
:shipit:

PR #309, opened 2022-04-20
Add account path

hansenreid:
Adds the GET Account path which returns a single account.

Will replace `/me` account info eventually.

We decided to rename the current `Account.yml` schema to `UserAccount` since we did not want to calculate `isDefault` and `isActive` for every call to this new path and does not make since outside of the context of a user.

We can rename the plural form `Accounts.yml` in the list Accounts story.


src/unity/paths/accounts_accountId.yml
```diff
@@ -24,7 +50,7 @@ patch:
       content:
         application/json:
           schema:
-            $ref: "../schemas/Account.yml"
+            $ref: "../schemas/UserAccount.yml"
```
randycoulman, line 53: I realize that this might be a potential breaking change in the UI, but would it be better to have this stay as the new `Account` schema for consistency instead?

It would mean we could avoid doing extra work in the `PATCH` controller.

On further thought, renaming schemas is already a "breaking" UI change, so maybe we could do this at the same time?

If you do make this change, can you add to the A/C of #6205 to remind us to actually do the simplification there?

src/unity/schemas/Accounts.yml
```diff
@@ -1,3 +1,3 @@
 type: array
 items:
-  $ref: './Account.yml'
+  $ref: './UserAccount.yml'
```
randycoulman, line 3: I'm thinking we should also rename `Accounts` -> `UserAccounts` as you suggested in the PR description. I think that would be less confusing.

randycoulman (approved):
This LGTM! A couple of suggestions below.

src/unity/paths/accounts_accountId.yml
hansenreid: Are you saying that we should update the Quartz controller in this story as we make the change, or as part of #6205?

randycoulman: I was thinking as part of #6205, but that might not make sense because the controller won't be returning the `type` and `billingProvider` fields yet.

I think that continuing to return the (soon to be) extra `isActive` and `isDefault` fields is fine until we do the controller work, but not returning the other two fields might confuse someone.

That said, I now see below that those two fields are not marked as `required` so maybe it's OK?

hansenreid: It seems that this change doesn't fit well here or in #6205, why don't I make a separate story to update the PATCH call in the controller and openapi to use the new `Account` schema?

randycoulman: Works for me.

hansenreid:
Updated `Accounts` to `UserAccounts` to avoid confusion.
Added new story to update PATCH endpoint to use `Account` schema: https://github.com/influxdata/quartz/issues/6276

I will make the corresponding UI change for the rename shortly after this is merged.

randycoulman (approved):
LGTM!

PR #312, opened 2022-04-21
chore: Add Profile path

hansenreid:
Adds the profile path that returns user profile information


hansenreid:
> Is this for #6142? If so, is there a reason we're not using `/users/:user_idpe_id` as the path and returning a `User` schema?

Yes it is for #6142. My mistake for not actually reading the last line of the story. Will update this accordingly.

randycoulman:
In retrospect, we could have made that clearer in the story. Sorry about that. We were going for consistency with `GET /accounts/:id` and `GET /orgs/:idpe_id`.

hansenreid:
Updated the path to be `users/{useridpeId}`. Also renamed the existing `User` Schema to `OperatorUser`. This type is not referenced directly by the UI, so no UI change is needed. (Tested by pushing a branch up with the SHA from the latest commit and all tests passed)

src/unity/schemas/OperatorUser.yml
```diff
@@ -0,0 +1,23 @@
...
+  idpeId:
+    type: string
+    description: user's id in idpe
+  lastName:
+    type: string
+    description: user's last name
+  operator:
+    type: boolean
+    description: whether or not the user is an operator
+  sfdcContactId:
```
randycoulman, line 20: Wow! I'm pretty sure this field doesn't exist any more, but that's out of scope for this PR.

randycoulman (approved):
LGTM!

src/unity/schemas/OperatorUser.yml
hansenreid: It is actually still being referenced by the UI

hansenreid: https://github.com/influxdata/ui/blob/2519757d1dcc2dc2163a313e784ed41a5722d7c2/src/operator/account/AccountGrid.tsx#L68

randycoulman: I'd be pretty surprised if it ever shows anything but `N/A`.

PR #313, opened 2022-04-21
chore: add two options to xygeom

blegesse-git:
see [ui#2651](https://github.com/influxdata/ui/issues/2651)

pr adds stepbefore and stepafter in XYGeom to expose options for line graphs


hoorayimhelping (approved):
LGTM (assuming you tested it against ui and it works) ;)

Issue #300, opened 2022-04-05
[ISSUE org deletion] Add org deletion endpoint to open api 300


StacieClark:
Deletion:
POST /api/v2private/unprovision/delete
{
orgID: "org ID"
}

Response: 204 OR a list of user ids or identities that were affected

PR #319, opened 2022-04-27
feat: provision delete org endpoint

wbaker85:
Closes https://github.com/influxdata/openapi/issues/300
See discussion here: https://github.com/influxdata/quartz/discussions/6039

Adds the API definition for the new `POST api/v2private/provision/delete` endpoint, which will be used to completely delete an organization from IDPE.


alejom99 (approved):
LGTM

PR #324, opened 2022-05-12
chore: deleting task runs returns 405 method not allowed in cloud

gavincabbage:
Changing the contract to communicate that cancelling task runs will not be allowed in Cloud once https://github.com/influxdata/idpe/issues/14307 is complete.

### DIFF
--- a/contracts/cloud.json
+++ b/contracts/cloud.json
@@ -5737,7 +5737,7 @@
         "tags": [
           "Tasks"
         ],
-        "summary": "Retrieve a single run for a task",
+        "summary": "Retrieve a single run for a task.",
         "parameters": [
           {
             "$ref": "#/components/parameters/TraceSpan"
@@ -5763,7 +5763,7 @@
         ],
         "responses": {
           "200": {
-            "description": "The run record",
+            "description": "Returns the run record.",
             "content": {
               "application/json": {
                 "schema": {
@@ -5773,7 +5773,7 @@
             }
           },
           "default": {
-            "description": "Unexpected error",
+            "description": "Unexpected error.",
             "content": {
               "application/json": {
                 "schema": {
@@ -5789,7 +5789,7 @@
         "tags": [
           "Tasks"
         ],
-        "summary": "Cancel a running task",
+        "summary": "Cancel a running task.\n\n#### InfluxDB Cloud\n\n  - Doesn't support this operation.\n",
         "parameters": [
           {
             "$ref": "#/components/parameters/TraceSpan"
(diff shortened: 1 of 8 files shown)


gavincabbage:
@glinton Is this the correct way to indicate a divergence like this?

glinton:
> @glinton Is this the correct way to indicate an OSS/Cloud divergence like this?

Not exactly, If that new response is only in cloud, you should split the diverged definition out of `common` and put it into `cloud` and `oss`; moving the file that was previously in `common`, to the directory of whichever platform isn't changing (`oss` it appears).

Or maybe the description you put is sufficient for the humans, and we don't worry about all that for a response code addition. I'll defer to @jstirnaman on that :smiley: 

```
Cancelling task runs is not supported in InfluxDB Cloud and will return a Method Not Allowed error
```

jstirnaman:
> > @glinton Is this the correct way to indicate an OSS/Cloud divergence like this?
> 
> Not exactly, If that new response is only in cloud, you should split the diverged definition out of `common` and put it into `cloud` and `oss`; 

@gavincabbage I'd agree with @glinton that defining the `path` separately for `cloud` and `oss` is technically more correct since the `operation` is absent in Cloud.
However, since splitting may lead to more divergence, you could follow [this example in `/write`](https://github.com/influxdata/openapi/blob/0c83679d7172b74c0f9863dda356f3ac26b73304/src/common/paths/write.yml#L222).
Tagging @lesam here since we were just discussing OSS-Cloud divergences.

jstirnaman:
>  However, since splitting may lead to more divergence, you could follow [this example in `/write`]
@gavincabbage And in `delete.description`, add something like this:
```
#### InfluxDB Cloud

- Doesn't support this operation.

lesam:
I prefer the approach in this PR vs splitting out the whole API definition.

The whole approach of having split API definitions doesn't work well for tooling that is supposed to be agnostic to Cloud/OSS usage.

Is there a plan to have OSS updated to reflect this API change?

glinton (approved):
I think the addition of a response code is fine to leave in common. It feels like we've done it ("not implemented" response) in the past for routes that are new to cloud but haven't landed in oss yet. One caveat to that being if the generated clients/consumers rely heavily on all responses being possible (again, this is an additive response so nothing should break).

gavincabbage:
Cool. Thanks for the discussion all. I will merge this as-is then.

> Is there a plan to have OSS updated to reflect this API change?

@lesam There is no plan to do that. This came about because cancellation of runs in progress will be made more difficult as a result of some changes we're making under the hood in Cloud, and metrics showed nobody was using this API successfully in Cloud anyway (one or two customers hit this endpoint and always receive a 404). I would rather add it back to Cloud in the someday™️ than remove it from OSS for no reason.

gavincabbage:
Request another review @glinton because I neglected to incorporate some of @jstirnaman's feedback re copying the `/write` example and calling out the differences in behavior between Cloud and OSS.

glinton (approved):
:tada:

Issue #299, opened 2022-04-05
[ISSUE org suspension] add suspension and unsuspension endpoints to open api 299


StacieClark:
mplement this endpont.
Setting the Org to suspended should be sufficent. Create an AUDIT message

POST /api/v2private/unprovision/suspend
{
orgID: "org ID"
}

Verifying with Quartz whether they want a Response: 204 OR a list of user ids or identities that were affected

unsuspension:
POST /api/v2private/unprovision/unsuspend
{
orgID: "org ID"
}

Verifying with Quartz whether they want a Response: 204 OR a list of user ids or identities that were affected

PR #325, opened 2022-05-16
feat: add provisioning suspend and unsuspend

wbaker85:
Closes https://github.com/influxdata/openapi/issues/299

### DIFF
--- a/contracts/priv/cloud-priv.yml
+++ b/contracts/priv/cloud-priv.yml
@@ -500,6 +500,62 @@ paths:
         default:
           $ref: '#/components/responses/ServerError'
           description: Unexpected error
+  /provision/suspend:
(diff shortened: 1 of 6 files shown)


StacieClark (commented):
If you define the request as a reusable request type, it makes it clearer that both of these APIs accept the same JSON.

wbaker85:
> If you define the request as a reusable request type, it makes it clearer that both of these APIs accept the same JSON.

Good idea...I was a little challenged by the name for it. I left it as `ProvisionSuspendRequest` and updated the description to indicate that it can suspend or unsuspend an org by id.

saraghds (approved):

StacieClark (approved):

PR #330, opened 2022-05-26
Add legacy query

candrewlee14:
Adds the 1.x compatibility `/query` endpoint.

- https://github.com/influxdata/influx-cli/pull/386 depends on this.

Because of `common.yml`'s server url prefix `/api/v2`, the raw `/query` path could not be added (and it would be duplicate path with the `/api/v2/query`. This led us to decide to make a new `legacy` folder, which holds the common legacy endpoints and resources with a blank server url prefix. This API gets injected into the final refs for cloud and oss through the `reference.sh` script.


src/common/paths/legacy_query.yml
```diff
@@ -0,0 +1,114 @@
+get:
```
jstirnaman, line 1: Do you mean for this to be included in the contracts? If so,
you need to set the PATH in https://github.com/influxdata/openapi/blob/master/src/common/_paths.yml, regenerate the contracts, and push them up.

jstirnaman: Will the path be `/legacy/query`, similar to https://github.com/influxdata/openapi/blob/a9df1ff92683aedf316bdb2650e146e2bdc7e6c3/src/oss.yml#L49?

jstirnaman: I'm going to PR some suggestions to your branch and then I'll update based on your replies.

contracts/oss.yml
```diff
@@ -6531,187 +6539,6 @@ paths:
         default:
           description: Unexpected error
           $ref: '#/components/responses/GeneralServerError'
-  /legacy/authorizations:
```
glinton, line 6534: I'm concerned that consumers of `contracts/oss.yml` will be missing this route (still found in ref/oss)

glinton (commented):
Another concern is that `/legacy/authorizations` now appears in ref/cloud, but i understood it doesn't actually exist in cloud. If that's the case, it may be better to keep the definitions where you put them, but don't reference them in `src/legacy.yml` and put them back into `src/oss.yml`. Thoughts?

candrewlee14:
> Another concern is that `/legacy/authorizations` now appears in ref/cloud, but i understood it doesn't actually exist in cloud. If that's the case, it may be better to keep the definitions where you put them, but don't reference them in `src/legacy.yml` and put them back into `src/oss.yml`. Thoughts?

Pinging @DStrand1. He says that `/legacy/authorizations` is also in cloud as well.

glinton (approved):
I realized it's not a loss to have these routes removed from `contracts/cloud` as it's an overall move towards modularity (a main goal of this repo), `ref` is intended to be a more "all-encompassing" contract, iirc eventually we'd like to break parts of oss/cloud out into similar service style contracts

src/oss.yml
```diff
@@ -127,10 +115,6 @@ components:
       $ref: "./common/schemas/Authorization.yml"
     AuthorizationPostRequest:
       $ref: "./common/schemas/AuthorizationPostRequest.yml"
-    LegacyAuthorizationPostRequest:
-      $ref: "./oss/schemas/LegacyAuthorizationPostRequest.yml"
-    Authorizations:
-      $ref: "./common/schemas/Authorizations.yml"
```
glinton, line 133: I believe this needs put back in to resolve @bednar's [last comment](https://github.com/influxdata/openapi/pull/330#discussion_r886341763)

candrewlee14: Yep, looks like I had deleted that and didn't notice because I saw that `Authorization` (no s) was still there.
Adding it back in.

contracts/oss.yml
```diff
@@ -6427,7 +6427,15 @@ paths:
...
-                $ref: '#/components/schemas/Authorizations'
+                type: object
+                properties:
+                  links:
+                    readOnly: true
+                    $ref: '#/components/schemas/Links'
+                  authorizations:
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/Authorization'
```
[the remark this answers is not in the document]
candrewlee14: I've added a new PR to add that `Authorizations` component which I had mistakenly removed, not noticing it wasn't a duplicate of `Authorization` (no s): #333 
I believe that should solve this for you, let me know. Sorry about that, thanks for catching it!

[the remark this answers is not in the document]
bednar: Thanks a lot @candrewlee14

PR #331, opened 2022-05-26
Docs/add legacy query

jstirnaman:
- Add `/legacy/query` to oss contracts.
- Revise descriptions, following https://github.com/influxdata/openapi/blob/master/docs/templates/pathTemplate.yml

### DIFF
--- a/src/common/paths/legacy_query.yml
+++ b/src/common/paths/legacy_query.yml
@@ -3,6 +3,7 @@ get:
   tags:
     - Legacy Query
   summary: Query with the 1.x compatibility API
+  description: Queries InfluxDB using InfluxQL.
   parameters:
     - $ref: "../parameters/TraceSpan.yml"
     - in: header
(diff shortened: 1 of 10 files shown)


src/common/paths/legacy_query.yml
```diff
@@ -2,14 +2,18 @@ get:
   operationId: GetLegacyQuery
```
candrewlee14, line 2: This file with all the wording changes is good.

src/oss.yml
```diff
@@ -166,6 +166,10 @@ paths:
     servers:
       - url: "/private"
     $ref: "./oss/paths/legacy_authorizations_authID_password.yml"
+  /legacy/query:
```
candrewlee14, line 169: The query path needs to be `/query` with no `/legacy/` or `/api/v2/` in front.
Since this would be a duplicate path as the `/query` under `/api/v2`, we may need to add `/api/v2` to all the paths manually to avoid the path collision.

Are the other changes in files here autogenerated?

candrewlee14 (changes requested):

PR #333, opened 2022-06-01
Fix missing OSS Authorizations component

candrewlee14:
When moving components over to legacy in #330, I mistakenly left out `Authorizations` in `oss.yml` (which would be a breaking change). This PR adds that component back.


glinton (approved):
:tada:

PR #334, opened 2022-06-02
chore(ecommerce): Camel case `billing_provider` in `Account.yml`

abshierjoel:
While pairing on some work with @wdoconnell we noticed `billing_provider` was snake-cased in `unity`. This changes the casing to lower camel case.


djcarpe (approved):

wdoconnell:
Thank you for fixing this @abshierjoel !

PR #336, opened 2022-06-02
ScriptHTTPResponseData as binary format

Xenonic:
This PR prevents immediate parsing of script invoke responses in memory (which may be massive) in the CLI, and instead lets the response be streamed on demand. This removes the overrides necessary in influxdata/influx-cli#387.


glinton (commented):
run `make generate-all` and push again

Xenonic:
had to manually fix up the generated files since `make generate-all` is failing to find `generate.sh`.

glinton (approved):

PR #338, opened 2022-06-06
Update the OEM api /orgs endpoint to reflect Quartz

eatondustin1:
We added the name field to the OEM orgs response in Quarz for the Multi-Org effort, and this updates the spec to match that. This also updates the spec for the other /orgs endpoints since they didn't match the existing behavior.

### DIFF
--- a/src/quartz/paths/org.yml
+++ b/src/quartz/paths/org.yml
@@ -44,7 +44,7 @@ delete:
       content:
         application/json:
           schema:
-            $ref: "../schemas/Organization.yml"
+            type: string
     "401":
       description: Unauthorized bearer token
       $ref: "../../common/responses/ServerError.yml"
--- a/src/quartz/schemas/Organization.yml
+++ b/src/quartz/schemas/Organization.yml
@@ -8,7 +8,7 @@ properties:
     description: URL for talking to a specific cluster
     readOnly: true
     type: string
-  orgName:
+  name:
     description: name of the organization
     type: string
   region:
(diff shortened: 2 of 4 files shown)


src/quartz/paths/org.yml
```diff
@@ -44,7 +44,7 @@ delete:
       content:
         application/json:
           schema:
-            $ref: "../schemas/Organization.yml"
+            type: string
```
eatondustin1, line 47: Grace and I decided to update the spec to match the existing Quartz behavior instead of updating Quartz itself. @randycoulman does this approach seem reasonable? I'm going to add a topic to the next tech discussion for what we think the command endpoints should return by default (entity being acted on vs success code), but wanted to at least make them match for now.

randycoulman: This looks like a bug in the spec; we shouldn't be returning the deleted org, IMO.

src/quartz/schemas/OrganizationWithToken.yml
```diff
@@ -6,6 +6,10 @@ allOf:
         description: Authentication token to manage the organization and its resources in IDPE
         readOnly: true
         type: string
+      userName:
```
randycoulman, line 9:
```suggestion
      name:
```

The OEM API knows nothing about users.

src/quartz/schemas/Organization.yml
```diff
@@ -8,6 +8,9 @@ properties:
     description: URL for talking to a specific cluster
     readOnly: true
     type: string
+  orgName:
```
randycoulman, line 11:
```suggestion
  name:
```

We're inside an `Organization` schema, so the `org` prefix is redundant.

randycoulman (approved):
Some suggestions below, but overall looks good.

randycoulman (approved):

PR #344, opened 2022-06-17
feat: add operator convert account to contract endpoint

eatondustin1:
This adds an endpoint to the unity API to convert and account to contract.


src/unity/paths/operator_accounts_accountId_convert.yml
```diff
@@ -0,0 +1,36 @@
+patch:
+  operationId: ConvertAccountById
+  tags:
+    - Accounts
+  summary: Convert an account to contract
+  requestBody:
+    description: The contract start date
```
randycoulman, line 7: Can/should this mention what the accepted format is? I'm assuming YYYY-MM-DD, but it would probably help the reader/consumer of the spec to know that for sure.

src/unity/schemas/OperatorAccountConvertRequest.yml
```diff
@@ -0,0 +1,7 @@
+properties:
+  contractStartDate:
+    type: string
+    format: date-time
+    description: the date the contract starts
```
randycoulman, line 5: Do we need a separate schema defined for this request? If so, then the `description` above should be generic, and the comment about the date and format should go here instead.

randycoulman (approved):
This looks good to me. A couple of little cleanups - suggested, but not required.

src/unity/schemas/OperatorAccountConvertRequest.yml
eatondustin1: I'm not sure if we need it, but it seems consistent with how we spec'ed  the "set default account" endpoint, and it makes maintenance easier in case we allow more "conversion" types/data. Good call on the descriptions, I'll make the change.

PR #346, opened 2022-06-22
fix: add 401 status code to idpe /me

wdoconnell:
IDPE's `/me` endpoint (`GetMe`) is able return a `401` status, but it is not reflected in the current path for `/me` (which reflects status codes `200` or `500` only). 

/me endpoint returning a `401` unauthorized status code
Screen Shot 2022-06-22 at 4 45 16 PM

This adds `401` as a valid status code  so that we can acknowledge it in the API layer of the UI.

### DIFF
--- a/src/common/paths/me.yml
+++ b/src/common/paths/me.yml
@@ -4,17 +4,17 @@ get:
     - Users
   summary: Retrieve the currently authenticated user
   parameters:
-    - $ref: '../parameters/TraceSpan.yml'
+    - $ref: "../parameters/TraceSpan.yml"
   responses:
-    '200':
+    "200":
       description: Success. The response body contains the currently authenticated user.
       content:
         application/json:
           schema:
-            $ref: '../schemas/UserResponse.yml'
-    '401':
-      $ref: '../responses/AuthorizationError.yml'
-    '500':
-      $ref: '../responses/InternalServerError.yml'
+            $ref: "../schemas/UserResponse.yml"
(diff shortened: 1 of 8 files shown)


src/common/paths/me.yml
```diff
@@ -12,6 +12,8 @@ get:
         application/json:
           schema:
             $ref: "../schemas/UserResponse.yml"
+    "401":
+      $ref: "../responses/AuthorizationError.yml"
```
wbaker85, line 16: Seems like just about everywhere else for `401` we use the code `"../../responses/ServerError.yml"`. `AuthorizationError.yml` seems to have specific language relating to authorization tokens. Authorization tokens can be used to hit this endpoint, so I guess that's fine, although it is mostly be used for browser sessions as far as I understand it. Might be something to keep in mind if any of the generated clients/docs are acting strange.

wbaker85 (approved):

wbaker85:
Reviewed for the services team since @SimonRichardson is OOO for the rest of this week and next.

src/common/paths/me.yml
wdoconnell: Thanks @wbaker85. We'll keep an eye out. For now I think we're okay with using the more specific `AuthorizationError` unless it concerns you. FYI, we're adding this so that we can distinguish the `4xx` and `5xx` errors at login so that we can implement a retry pattern with backoff for certain non-`401` responses.

PR #347, opened 2022-06-22
fix: over cardinality status code

jstirnaman:
- If request exceeds cardinality limit, Cloud returns 429 instead of 503.

https://github.com/influxdata/DAR/issues/304

### DIFF
--- a/contracts/cloud.json
+++ b/contracts/cloud.json
@@ -1741,7 +1741,7 @@
             }
           },
           "429": {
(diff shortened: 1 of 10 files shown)


abalone23 (commented):
A Retry-After header will not be returned for a cardinality rate limit.

jstirnaman:
> A Retry-After header will not be returned for a cardinality rate limit.

@abalone23 Please review again. Thanks!

glinton (approved):

PR #349, opened 2022-06-23
feat(tasks): add API changes for tasks to account for scripts addition

yquansah:
This PR proposes changes to the openapi spec to account for the feature for tasks to invoke scripts at runtime. There are changes to the following:

- `GET /api/v2/tasks`
- `POST /api/v2/tasks`
- `PATCH /api/v2/tasks/{id}`


brettbuddin (approved):

mhilton (changes requested):
This looks like it is updating the OSS documentation, I don't think that's correct. OSS doesn't support these changes.

yquansah:
Alright looks like I am doing something terribly wrong 😅

src/cloud.yml
```diff
@@ -38,6 +38,8 @@ paths:
     $ref: './cloud/paths/dashboards.yml'
   /tasks:
     $ref: './cloud/paths/tasks.yml'
+  "/tasks/{taskID}":
```
glinton, line 41: For other reviewers, this was necessary because it referenced the task schema.

glinton (approved):
Excellent, this pr is a fantastic example of what needs to happen when oss/cloud apis drift

src/common/_paths.yml
```diff
@@ -116,8 +116,6 @@
     $ref: "./common/paths/templates_apply.yml"
   /templates/export:
     $ref: "./common/paths/templates_export.yml"
-  "/tasks/{taskID}":
-    $ref: "./common/paths/tasks_taskID.yml"
```
glinton, line 120: This should have been added to `src/oss.yml`, maybe a file save was forgotten

contracts/oss-diff.yml
```diff
@@ -5793,6 +5750,162 @@ components:
           type: integer
           format: int64
           minimum: 0
+    Tasks:
+      type: object
+      properties:
+        links:
+          readOnly: true
+          $ref: '#/paths/~1dashboards/get/responses/200/content/application~1json/schema/properties/links'
```
yquansah, line 5758: I am a little unsure about this bit here @glinton. If there are other things I would need to do to fix this up.

glinton: Ah yes, the `-diff` files are nice to have, but are in no way a prioirty and haven't gotten any love. they aren't used by the ui and their terrible generator `oats` as yet, so we ignore those inline references. The solution would be to go through and add an entry for the schemas, responses, etc that is being used. Could be done simply/hacky in a similar pattern to the `_schemas.yml` `bash` almost template thing we do now, but isn't necessary.

Xenonic (approved):
LGTM!

contracts/oss-diff.yml
```diff
@@ -4518,11 +4518,11 @@ paths:
                     type: object
                     properties:
                       next:
-                        $ref: '#/components/schemas/ScraperTargetResponse/allOf/1/properties/links/properties/self'
+                        $ref: '#/components/schemas/Task/properties/links/properties/self'
```
mhilton, line 4521: It seems odd that this has changed, but I expect there is only really one definition of "self" required.

mhilton (approved):

gavincabbage:
Taking notes on proposed changes while talking to @yquansah:

- Will add clarity on use of either `flux` or `scriptID`. We will use `oneOf` if possible in our openAPI version, or just add details to the descriptions of `flux`, `scriptID` and `scriptParameters` if not. 
- Add a description to the "Create a new task" endpoint and probably the update endpoint as well that calls our the ability to create a script with Flux or by referencing a script and that one is not compatible with the other. 
- Open an issue to link to the not-yet-existing docs page in the description mentioned above.

jstirnaman (approved):
I'll port my changes from https://github.com/influxdata/openapi/pull/432 to the new `src` files.

jstirnaman:
@yquansah @gavincabbage
The current minimum for limit is `1`.
What does `limit=-1` do in https://github.com/influxdata/openapi/blob/add57d884c3827e47b21324e963c6dbc99eae9ab/src/cloud/paths/tasks.yml#L46?

yquansah:
@jstirnaman good question. I am not particularly sure of what that `minimum` will do. I can dig around to see

jstirnaman:
> > * Add a description to the "Create a new task" endpoint and probably the update endpoint as well that calls our the ability to create a script with Flux or by referencing a script and that one is not compatible with the other.
> 
> @yquansah @gavincabbage While not ideal having to rely on human parsing, you can find somewhat of a pattern for this in https://github.com/influxdata/openapi/blob/master/src/oss/paths/debug_pprof_allocs.yml#L33-52

I'm working on this now, fyi.

glinton:
I believe if `limit=-1` is passed in, it doesn't paginate tasks, but attempts to return all of them at once to the user (intended to be used with `type=basic` to avoid returning individual task flux in the response).

jstirnaman:
> I believe if `limit=-1` is passed in, it doesn't paginate tasks, but attempts to return all of them at once to the user (intended to be used with `type=basic` to avoid returning individual task flux in the response).

Will `type=basic` include the `scriptID` and `scriptParameters` or send them empty as well?

glinton:
> Will `type=basic` include the `scriptID` and `scriptParameters` or send them empty as well?

I'm not certain, but those fields will likely be sent back as well, if they are set. I know they are tagged `omitempty` which means the field will not be returned if it's not set (something easily changeable if desired)

yquansah:
I will fix conflicts on this PR and then we can merge when necessary.

PR #353, opened 2022-06-27
feat: add instance resource type for oss

jeffreyssmith2nd:
OSS is introducing the concept of an `instance` type that allows an approved user full permissions to everything in the instance (including adding additional orgs, etc).

### DIFF
--- a/contracts/legacy.yml
+++ b/contracts/legacy.yml
@@ -628,6 +628,7 @@ components:
                 - annotations
                 - remotes
                 - replications
+                - instance
             id:
               type: string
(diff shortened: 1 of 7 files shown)


jeffreyssmith2nd:
> The change you made in `common` Resources file to add instance type affects refs for both cloud and OSS.
> Is this going to be a cloud feature or only OSS, and if it's only OSS, should it be moved from `common` to `oss`?

It is OSS only. Remotes and replications are included in this file even though they are OSS only, so I followed the pattern there. My quick research showed there's no way to merge too separately defined enums so this is the best we can do atm.

candrewlee14 (approved):
LGTM

DStrand1 (approved):

PR #357, opened 2022-06-29
Chore/api reference delete

sunbryely-influxdata:
Update Delete API Reference Docs 

https://github.com/influxdata/docs-v2/projects/9#card-83430069

- Added cURL Command
- Added 204 Empty Response Sample

image

### DIFF
--- a/src/common/paths/delete.yml
+++ b/src/common/paths/delete.yml
@@ -58,9 +58,9 @@ post:
       **delete predicate expression** in the `predicate` property of the request body.
       If you don't pass a `predicate`, InfluxDB deletes all data with timestamps
       in the specified time range.
-      
+
       #### Related guides
-      
+
       - [Delete data]({{% INFLUXDB_DOCS_URL %}}/write-data/delete-data/).
       - Learn how to use [delete predicate syntax]({{% INFLUXDB_DOCS_URL %}}/reference/syntax/delete-predicate/).
     required: true
--- a/src/common/paths/delete.yml
+++ b/src/common/paths/delete.yml
@@ -42,16 +42,14 @@ post:
     - lang: Shell
       label: cURL
       source: |
-        
-        curl --request POST http://localhost:8086/api/v2/delete?org=example-org&bucket=example-bucket \
-        --header 'Authorization: Token YOUR_API_TOKEN' \
-        --header 'Content-Type: application/json' \
-        --data '{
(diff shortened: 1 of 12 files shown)


src/common/paths/delete.yml
```diff
@@ -47,15 +59,14 @@ post:
       If you don't pass a `predicate`, InfluxDB deletes all data with timestamps
       in the specified time range.
 
-      #### Related guides
```
sunbryely-influxdata, line 50: This is written on line 37. I removed it, but if we want to keep a `Related guides` section here, I can keep it in place.

src/common/paths/delete.yml
```diff
@@ -128,6 +139,9 @@ post:
         #### InfluxDB OSS
 
         - Deleted the data.
+      content:
```
sunbryely-influxdata, line 142: 204 Successful Response was not showing before, so I added an empty response since that is technically what is being returned

src/common/paths/delete.yml
```diff
@@ -38,6 +38,18 @@ post:
     - Learn how to use [delete predicate syntax]({{% INFLUXDB_DOCS_URL %}}/reference/syntax/delete-predicate/).
     - Learn how InfluxDB handles [deleted tags](https://docs.influxdata.com/flux/v0.x/stdlib/influxdata/influxdb/schema/measurementtagkeys/)
       and [deleted fields](https://docs.influxdata.com/flux/v0.x/stdlib/influxdata/influxdb/schema/measurementfieldkeys/).
+  x-codeSamples:
+    - lang: Shell
+      label: cURL
+      source: |
+        curl --request POST {INFLUX_URL}/api/v2/delete?org={INFLUX_ORG}&bucket={INFLUX_BUCKET} \
```
sunbryely-influxdata, line 45: @jstirnaman should i remove the braces around the placeholders?

src/common/paths/delete.yml
jstirnaman: > 

@sunbryely-influxdata Researched this a bit and didn't any good solution for displaying '204' in the response preview pane while also keeping the response content-free. Seems to be a common problem/wish across other OpenAPI tools. Rapidoc appears to behave similarly.

Once we beef up the spec, then we can replace Redoc and build the UI behavior that we want.

jstirnaman (approved):
Looks good!

noramullen1 (approved):

Issue #355, opened 2022-06-29
OSS header errors don't agree with spec


jstirnaman:
`src/common/paths/query_analyze.yml` contains

```yaml
    default:
      description: Internal server error
      headers:
        X-Influx-Error:
          description: A string that describes the problem.
          schema:
            type: string
        X-Influx-Reference:
          description: The numeric reference code for the error type.
          schema:
            type: integer
```

but I don't see either of those error headers returned by OSS; I only see `X-Platform-Error-Code`.

We need to compare this with Cloud and then fix the spec or document the differences.
For now, I've added it as an additional header:

```yaml
    default:
      description: Internal server error
      headers:
        X-Platform-Error-Code:
          description: The reason for the error.
          schema:
            type: string
            example: "internal error"
        X-Influx-Error:
          description: A string that describes the problem.
          schema:
            type: string
        X-Influx-Reference:
          description: The numeric reference code for the error type.
          schema:
            type: integer
```

PR #359, opened 2022-06-30
docs: Description and examples for query/analyze:

### LINKED ISSUE #356: api-reference: Analyze a Query
Test and improve spec for all query endpoints.

Related:
https://github.com/influxdata/openapi/issues/355

jstirnaman:
- Improve description.
- Add request and response examples.
- Add code sample.

Closes #356
Opened a related issue #355 for response error headers.

### DIFF
--- a/contracts/cloud.json
+++ b/contracts/cloud.json
@@ -3262,6 +3262,7 @@
           "Query"
         ],
         "summary": "Analyze a Flux query",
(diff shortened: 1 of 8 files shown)


jstirnaman:
@lesam Do you happen to know the truth about headers for error responses?
My results for OSS didn't match the spec, so I added `X-Platform-Error-Code` and left the others in https://github.com/influxdata/openapi/pull/359/files#diff-37d74b6840762ed4a624722f21d65adb2cbf0194a650534c2490f655a4312ebcR115-R127.

I created #355 for follow-up.

sunbryely-influxdata (approved):

glinton:
> Do you happen to know the truth about headers for error responses?

Not exactly, just that there is one default error handler that sets `X-Platform-Error-Code` but using it isn't exactly standard, so that header may not always come through, but a nearly equivalent value more likely always comes through as the response code. In other words, it may enrich the response, but not having it isn't the end of the world.

There is a comment in the code: `We're no longer using X-Influx-Error and X-Influx-Reference.`, so I'd imagine removing those from the spec is a move in the right direction.

glinton (approved):

PR #360, opened 2022-06-30
Docs/api reference tasks

jstirnaman:
- Establish a general pattern for list query param descriptions.
- Improve descriptions for list and create.
- Add request and response examples.
- Add code samples.

### DIFF
--- a/src/common/paths/tasks.yml
+++ b/src/common/paths/tasks.yml
@@ -5,7 +5,7 @@ get:
   summary: List tasks
   description: |
     Retrieves a list of [tasks]({{% INFLUXDB_DOCS_URL %}}/process-data/).
-    
+
     To limit which tasks are returned, pass query parameters in your request.
     If no query parameters are passed, InfluxDB returns all tasks up to the default `limit`.
   parameters:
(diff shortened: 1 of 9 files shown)


glinton:
The `type` parameter exists when querying tasks and provided a hacky method (`type=basic`) to list tasks without returning their flux (was bogging down the ui). I know the ui hasn't adopted this and still fetches all tasks and their flux. We can likely revisit this filtering to be more idiomatic at a later date.

src/common/paths/tasks.yml
```diff
@@ -2,53 +2,75 @@ get:
...
+        The minimum is `1`, the maximum is `500`, and the default is `100`.
     - in: query
       name: type
-      description: Type of task, unset by default.
+      description: |
+        Task type (`basic` or `system`).
+        If `system`, returns additional metadata fields for tasks; otherwise, returns **basic** task fields.
       required: false
       schema:
         default: ""
```
glinton, line 76: iirc, the server defaults to `system`

glinton (approved):

sunbryely-influxdata (approved):

glinton:
> `type` property in the `Task` schema. Is it related? Not fully implemented?

The `type` property in the schema is implemented on the internal task api as well as on the check response externally (internally a task and a check is the same thing, differentiated by type)? Since we have a dedicated schema for checks, I believe we can safely remove the type property from the task schema definition.

PR #429, opened 2022-07-11
feat: add status route to get status for all subscriptions

subirjolly:
Related: https://github.com/influxdata/data-acquisition/issues/329

Add `/broker/subs/status` route to fetch status for all subscriptions.


src/svc/nifid/schemas/SubscriptionStatusArray.yml
```diff
@@ -0,0 +1,5 @@
+properties:
+  subscriptions:
+    type: array
+    items:
+      $ref: "./SubscriptionStatus.yml"
```
helenosheaa, line 5: we'll need to extend this definition as it currently just has a property of isActive

helenosheaa (changes requested):

subirjolly:
@helenosheaa Thoughts on this now? There's a side-effect. As I changed the SubscriptionStatus and made the payload a separate entity(anonymous), this means now the `data` for `put` is not of type `SubscriptionStatus`.
```
export interface PutBrokerSubsStatusParams {
  id: string

  data: {
    isActive?: boolean
  }
}
```

src/svc-nifid.yml
```diff
@@ -11,6 +11,8 @@ paths:
     $ref: './svc/nifid/paths/broker_subs_id.yml'
   /broker/subs/{id}/status:
     $ref: './svc/nifid/paths/broker_subs_id_status.yml'
+  /broker/subs/statuses:
+    $ref: './svc/nifid/paths/broker_subs_status.yml'
```
helenosheaa, line 15: I think you need to add the new schemas and request body change in this file then it should fix the auto generated ones in contracts

helenosheaa (commented):
looking good, left a few comments

contracts/svc/nifid.yml
```diff
@@ -181,7 +202,9 @@ components:
       content:
         application/json:
           schema:
-            $ref: '#/components/schemas/SubscriptionStatus'
+            properties:
```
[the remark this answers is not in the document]
helenosheaa: I think it's what I said below about updated that svc-nifi file

contracts/svc/nifid.yml
```diff
@@ -166,6 +166,27 @@ paths:
...
+        '200':
+          description: returns Status of all Subscription records in org
+          content:
+            application/json:
+              schema:
+                properties:
+                  subscriptions:
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/SubscriptionStatus'
```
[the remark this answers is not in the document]
subirjolly: this is getting auto-generated. so, i can't change this.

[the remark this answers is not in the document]
helenosheaa: I think if it was defined differently in the other file it would be autogenerated differently more like SubscriptionArray but no big deal.

src/svc-nifid.yml
```diff
@@ -32,6 +34,8 @@ components:
       $ref: './svc/nifid/schemas/Subscriptions.yml'
     SubscriptionStatus:
       $ref: './svc/nifid/schemas/SubscriptionStatus.yml'
+    SubscriptionRunStatus:
```
helenosheaa, line 37:
```
 SubscriptionStatusArray:
 $ref: './svc/nifid/schemas/SubscriptionStatusArray.yml'
```

 should be under these schemas as well

helenosheaa (changes requested):
Looking good! Just one schema missing in the svc-nifid file then should be good

src/svc-nifid.yml
subirjolly: yes, good catch! I was.. testing, reverting, blah, blah... repeat.. did that and reverted it(apparently).

Thank you!

helenosheaa (approved):
looks good!

PR #430, opened 2022-07-12
feat(ecommerce): Add user response to organization invite

abshierjoel:
When a user is invited to an organization, but they already belong to the account of that organization, they will instead be automatically added to the organization without sending an invite. In this case, we'll return back their user to be added to the UI.


eatondustin1 (approved):
Look good! We should probably go ahead and open a PR in the UI just to update the SHA for the generated client.

PR #431, opened 2022-07-12
Chore/api reference query suggestions

sunbryely-influxdata:
https://github.com/influxdata/openapi/issues/395

`v2/query/suggestions`


src/common/paths/query_suggestions.yml
```diff
@@ -2,18 +2,64 @@ get:
...
+    functions that are not meant to be used while building a query such as
+    [`_sortLimit()`](https://github.com/influxdata/flux/blob/a768568a4cbeb299433ee31a059e599d290fcfbf/libflux/flux/benches/everything.flux#L2247)
+
+    #### Limitations
+
+    - The endpoint does not return all of the functions in the Flux standard
+    library. It only returns query functions.
+
+    - Using `/api/v2/query/suggestions/` with cURL will result in a HTTP `301
+    Moved Permanently` status code. Please use `/api/v2/query/suggestions`.
```
sanderson, line 20: This took me a second to process. Maybe call out the fact that the trailing slash is what causes the problem.

src/common/paths/query_suggestions.yml
```diff
@@ -2,19 +2,65 @@ get:
...
         application/json:
           schema:
             $ref: "../schemas/Error.yml"
+  x-codeSamples:
+    - lang: Shell
+      label: cURL
+      source: |
+        curl --request GET 'INFLUX_URL/api/v2/query/suggestions' \
+          --header 'Accept: application/json \
+          --header 'Authorization: Token INFLUX_TOKEN'
```
sanderson, line 66:
```suggestion
        curl --request GET 'INFLUX_URL/api/v2/query/suggestions' \
          --header 'Accept: application/json' \
          --header 'Authorization: Token INFLUX_TOKEN'
```

src/common/paths/query_suggestions.yml
```diff
@@ -2,19 +2,65 @@ get:
   operationId: GetQuerySuggestions
   tags:
     - Query
-  summary: Retrieve query suggestions
+  summary: Retrieve all query functions
+  description: |
+    Returns a list of all query functions.
```
sanderson, line 7: What makes these "query functions"? This is new nomenclature.

I feel like the original summary of this endpoint was correct. It returns a limited list of query/function suggestions. I'm just curious how this endpoint is even used. It seems like it's probably specific to the UI implementation to display functions in a specific context. I don't know what they used to determine which functions to return with this endpoint.

src/common/paths/query_suggestions.yml
```diff
@@ -2,19 +2,65 @@ get:
...
+        Use `/api/v2/query/suggestions`.
+      content:
+        text/html:
+          schema:
+            $ref: "../schemas/MovedPermanently.yml"
+          examples:
+            movedPermanently:
+              summary: Invalid url
+              description: |
+                The url has been permanently moved. Use `/api/v2/query/suggestions`.
```
sanderson, line 51:
```suggestion
                The URL has been permanently moved. Use `/api/v2/query/suggestions`.
```

src/common/schemas/MovedPermanently.yml
```diff
@@ -0,0 +1,5 @@
+  properties:
+    body:
+      readOnly: true
+      description: Response message with url of requested resource.
```
sanderson, line 4:
```suggestion
      description: Response message with URL of requested resource.
```

src/common/paths/query_suggestions.yml
```diff
@@ -2,19 +2,65 @@ get:
...
+        Moved Permanently.
+        InfluxData has moved the url of the endpoint.
+        Use `/api/v2/query/suggestions`.
+      content:
+        text/html:
+          schema:
+            $ref: "../schemas/MovedPermanently.yml"
+          examples:
+            movedPermanently:
+              summary: Invalid url
```
sanderson, line 49:
```suggestion
              summary: Invalid URL
```

src/common/paths/query_suggestions.yml
```diff
@@ -2,19 +2,65 @@ get:
...
       content:
         application/json:
           schema:
             $ref: "../schemas/FluxSuggestions.yml"
+          examples:
+            $ref: "../schemas/QuerySuggestionsSuccessResponse.yml"
+    "301":
+      description: |
+        Moved Permanently.
+        InfluxData has moved the url of the endpoint.
```
sanderson, line 41:
```suggestion
        InfluxData has moved the URL of the endpoint.
```

sanderson (changes requested):

jstirnaman (approved):
👍

src/common/paths/query_suggestions.yml
```diff
@@ -2,19 +2,72 @@ get:
...
+    [Flux function](https://docs.influxdata.com/flux/v0.x/stdlib/all-functions/)
+    name and parameters.
+
+    Use this endpoint to retrieve a list of Flux query suggestions used in the
+    InfluxDB Flux Query Builder. Helper function names have an underscore (`_`)
+    prefix and shouldn't be used  in queries--for example:
+
+    - **Not recommended**: Don't use the `_sortLimit` helper function directly.
+       - **Recommended**:  Use `top(n, columns=["_value"], tables=<-)` to sort
+        on a column and keep the top n records. `top` uses the `_sortLimit` helper function.
```
sanderson, line 17: Minor nit: It's not that they shouldn't be used in queries, it's just that they aren't meant to be used directly. They actually are used by other functions as primitives. I don't know if it's worth updating this verbiage, but just a thought. In that same vein, I don't know that the "Not recommended vs Recommended" is necessary.

sanderson (approved):

PR #434, opened 2022-07-13
Revert "✨ Add `users` to create OEM Org endpoint"

abshierjoel:
Reverts influxdata/openapi#305

We decided that we no longer need this behavior.


eatondustin1 (approved):
🔥

jstirnaman (approved):

PR #438, opened 2022-07-14
docs(tasks): task labels and error fixes.

### LINKED ISSUE #408: api-reference: Tasks - list all labels for a task

### LINKED ISSUE #409: api-reference: Tasks - add a label to a task

### LINKED ISSUE #410: api-reference: Tasks - delete a label from a task

jstirnaman:
- Improve descriptions for task labels.
- Improve query params description.
- Remove OSS org errors where they don't apply (define them in write, query, delete).

Closes #408
Closes #409
Closes #410

### DIFF
--- a/src/common/paths/write.yml
+++ b/src/common/paths/write.yml
@@ -205,20 +205,36 @@ post:
         - [How to check for write errors]({{% INFLUXDB_DOCS_URL %}}/write-data/troubleshoot/).
     "400":
       description: |
-        Bad request.
-        The line protocol data in the request is malformed.
+        Bad request. The response body contains detail about the error.
+
+        InfluxDB returns this error if the line protocol data in the request is malformed.
         The response body contains the first malformed line in the data.
         InfluxDB rejected the batch and did not write any data.
+
+        #### InfluxDB Cloud
+
+        - Returns this error for bucket schema conflicts.
+
+        #### InfluxDB OSS
+
+        - Returns this error if `org` or `orgID` does not match an organization.
+
       content:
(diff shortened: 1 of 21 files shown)


src/common/paths/write.yml
```diff
@@ -205,21 +205,37 @@ post:
...
+              summary: (Cloud) field type conflict thrown by an explicit bucket schema
               value: {
                "code": "invalid",
                "message": "partial write error (2 written): unable to parse 'air_sensor,service=S1,sensor=L1 temperature=\"90.5\",humidity=70.0 1632850122': schema: field type for field \"temperature\" not permitted by schema; got String but expected Float"
               }
+            orgNotFound:
+              summary: (OSS) organization not found
+              value: {
+                "code":"invalid",
+                "message":"failed to decode request body: organization not found"
```
glinton, line 237: This is another curious one, is the error saying that the organization wasn't found in the payload, or in the system. If the latter, this is an enumeration bug. (Maybe it's only in oss?)

glinton (approved):
Looks good, just some interesting responses that @influxdata/services-team may want to check out in the code.

src/common/paths/write.yml
jstirnaman: > This is another curious one, is the error saying that the organization wasn't found in the payload, or in the system. If the latter, this is an enumeration bug. (Maybe it's only in oss?)

The latter, when I pass an incorrect orgID-ish orgID. I assume it's when the token doesn't belong to that orgID?

PR #439, opened 2022-07-15
chore: passwords API on cloud not supported

wbaker85:
Closes https://github.com/influxdata/idpe/issues/15092

Changing passwords is not supported through the InfluxDB Cloud API, and this PR updates the API docs for that.

### DIFF
--- a/src/common/paths/me_password.yml
+++ b/src/common/paths/me_password.yml
@@ -6,7 +6,8 @@ put:
   description: |
     #### InfluxDB Cloud
 
-    - Doesn't support changing user passwords through the API.
+    InfluxDB Cloud does not support changing user passwords through the API.
+    Use the InfluxDB Cloud user interface to update your password.
   security:
     - BasicAuthentication: []
   parameters:
--- a/src/common/paths/me_password.yml
+++ b/src/common/paths/me_password.yml
@@ -3,23 +3,31 @@ put:
   tags:
     - Users
   summary: Update a password
+  description: |
+    #### InfluxDB Cloud
+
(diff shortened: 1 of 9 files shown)


src/common/paths/me_password.yml
```diff
@@ -3,23 +3,31 @@ put:
   tags:
     - Users
   summary: Update a password
+  description: |
+    #### InfluxDB Cloud
+
+    - Doesn't support changing user passwords through the API.
```
sanderson, line 9:
```suggestion
    InfluxDB Cloud does not support changing user passwords through the API.
    Use the InfluxDB Cloud user interface to update your password.
```

src/common/paths/users_userID_password.yml
```diff
@@ -4,10 +4,14 @@ post:
     - Security and access endpoints
     - Users
   summary: Update a password
+  description: |
+    #### InfluxDB Cloud
+
+    - Doesn't support changing user passwords through the API.
```
sanderson, line 10:
```suggestion
    InfluxDB Cloud does not support changing user passwords through the API.
    Use the InfluxDB Cloud user interface to update your password.
```

sanderson (changes requested):
Some minor suggestions and alternate calls to action.

src/common/paths/me_password.yml
wbaker85: Done, thanks!

sanderson (approved):

SimonRichardson (approved):

PR #445, opened 2022-07-20
chore: condensed tags when multiple were present to fix codegen

Xenonic:
This PR fixes the codegen used by the CLI tool, which broke in https://github.com/influxdata/openapi/pull/427. The root of the problem is when the generator sees an operation with multiple tags (eg. `Delete` and `Data I/O endpoints`), it will generate conflicting definitions in each tag's interface. To fix this, the tags have been condensed into the more general grouping tag, in the example case the `Data I/O endpoints` take would replace the `Delete` tag.

Note there is also a fix to additional broken codegen introduced in https://github.com/influxdata/openapi/pull/360, where the 400 status error type has been changed to `Error.yml` instead of the more specific `BadRequestError.yml`, which wouldn't compile. Affects `src/common/paths/tasks.yml`

### DIFF
--- a/contracts/cloud-diff.yml
+++ b/contracts/cloud-diff.yml
@@ -11,7 +11,7 @@ paths:
     get:
       operationId: GetUsers
       tags:
-        - Users
+        - Security and access endpoints
       summary: List all users
       parameters:
         - in: header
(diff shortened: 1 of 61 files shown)


jstirnaman (changes requested):
@Xenonic I'd prefer to not remove these tags from `src`. OpenAPI supports multiple tags so the problem is with whatever is consuming the contracts. That said, if we have to remove a tag, please keep the more specific tags (`Authorization`), and remove the "category" tags (`Security and access endpoints`). I can apply those downstream for docs.

Sorry, I guess I misunderstood what you meant by "codegen" in our Slack discussion.
I thought you meant whatever tool you used to consume the contracts and generate code.

PR #446, opened 2022-07-20
fix: removed incorrect schema tag in PostTasks 400 response to fix CLI codegen

Xenonic:
In `tasks.yml`, the 400 response of the `PostTasks` operation has an incorrect `content/application/schema` tag structure that no other responses use. This has now been flattened and the `description`/`examples` tags moved up.

### DIFF
--- a/src/cloud/paths/tasks.yml
+++ b/src/cloud/paths/tasks.yml
@@ -176,25 +176,22 @@ post:
 
           - Returns this error if the task doesn't contain one of _`flux`_ or _`scriptID`_.
           - Returns this error if the task contains _`flux`_ _and_ _`scriptID`_.
-      content:
-        application/json:
-          schema:
-            $ref: "../../common/responses/BadRequestError.yml"
-          examples:
-            fluxAndScriptError:
-              summary: The request body can't contain both flux and scriptID
-              value:
-                {
-                  "code": "invalid",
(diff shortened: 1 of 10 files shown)


jstirnaman (changes requested):
We want to be specific when documenting errors. Let's correct it or fix the consumer problem.

Xenonic:
@jstirnaman sorry that was a bit of a hacky fix when working on the CLI, latest fixes codegen and should have no impact on the docs.

src/cloud/paths/tasks.yml
```diff
@@ -176,25 +176,22 @@ post:
...
-                  "message": "failed to decode request: can not provide both scriptID and flux"
-                }
-            missingFluxError:
-              summary: The request body requires either flux or scriptID
-              value:
-                {
-                  "code": "invalid",
-                  "message": "failed to decode request: flux required"
-                }
+      $ref: "../../common/responses/BadRequestError.yml"
```
jstirnaman, line 179: @Xenonic Thanks for fixing this! I'm surprised OATS validation let it through. I made a similar mistake yesterday and OATS CI failed it.
FYI, I think `$ref` with siblings isn't technically valid OpenAPI, but it's a rule that we frequently break without problems (AFAIK).

jstirnaman (approved):

PR #447, opened 2022-07-21
clarify 400 on partial writes

kelseiv:
closes
https://github.com/influxdata/DAR/issues/312


src/common/paths/write.yml
```diff
@@ -208,8 +208,9 @@ post:
         Bad request. The response body contains detail about the error.
 
         InfluxDB returns this error if the line protocol data in the request is malformed.
-        The response body contains the first malformed line in the data.
-        InfluxDB rejected the batch and did not write any data.
+        The response body contains the first malformed line in the data, and indicates what was expected.
```
kelseiv, line 211: @sunbryely-influxdata, not sure if indicates what was expected is always true. If it's not, we could drop that, or say in some cases.

sunbryely-influxdata: ill do a bit more testing to see if we should drop that line

sunbryely-influxdata (approved):

src/common/paths/write.yml
kelseiv: thank you, @sunbryely-influxdata! cc @Joe-Influx

PR #448, opened 2022-07-21
chore: Add endpoint for getting orgs by account

hansenreid:
part of https://github.com/influxdata/quartz/issues/6604

Adds endpoint for returning a list of Organizations for the given account. Returns the same payload as `/orgs`.


eatondustin1 (approved):
👍

PR #449, opened 2022-07-22
chores: Add endpoint for setting account default org

gspletzer:
Part of Quartz [#6603](https://github.com/influxdata/quartz/issues/6603)

Add `PUT /api/v2private/accounts/:account_id/orgs/default` to unity. This will allow a default org to be set for any account, not just `active` accounts. It works similar to the unity path for allowing a user to set a default organization. 

Additionally specific errors were added to the spec for both `PUT /api/v2private/accounts/:account_id/orgs/default` and `PUT /api/v2private/orgs/default` to match the returns specified in Quartz.


src/unity.yml
```diff
@@ -17,6 +17,8 @@ paths:
     $ref: './unity/paths/accounts_accountId.yml'
   '/accounts/{accountId}/orgs':
     $ref: './unity/paths/accounts_accountId_orgs.yml'
+  '/accounts/{accountId}/orgs/default':
+    $ref: './unity/paths/accounts_org_default.yml'
```
eatondustin1, line 21: I think the file name should be `accounts_accountId_orgs_default` to match the naming standards, just basing it on the others. Just let me know if I'm wrong.

```suggestion
    $ref: './unity/paths/accounts_accountId_orgs_default.yml'
```

eatondustin1 (changes requested):
Just one requested change below, otherwise it looks good. Nice catch on the error types!

eatondustin1 (approved):
👍

PR #454, opened 2022-07-27
chore(billing): Remove Region from BillingInfo

distortia:
Removes the `region` from the `BillingInfo` Schema. The region is no longer needed in a multi-org/account world and instead of fetching the region for the active org, we simply removed it.


abshierjoel (approved):
🥳 🐱

PR #455, opened 2022-07-28
feat(unity): remove operator delete user from account endpoint

randycoulman:
The delete account user endpoint only works when there is a 1:1 mapping between accounts and organizations. With multi-org, that is about to change.

We will be revisiting user removal from orgs and accounts in a later phase of multi-account, so for now we are deleting this endpoint.

The UI code that was calling this endpoint was deleted in https://github.com/influxdata/ui/pull/5238.


eatondustin1 (approved):
🔥

hoorayimhelping (approved):

PR #462, opened 2022-08-03
feat: add virtual field to DBRP schema

candrewlee14:
Supports this auto-DBRP PR:
- https://github.com/influxdata/influxdb/pull/23606


src/common/schemas/DBRP.yml
```diff
@@ -19,6 +19,9 @@
     default:
       type: boolean
       description: Mapping represents the default retention policy for the database specified.
+    virtual:
+      type: boolean
```
glinton, line 23: Is this field added to the response in cloud as well or is this oss only?

candrewlee14: It's currently only OSS, but I think this auto-mapping DBRP feature is probably something that would be ported to cloud as well.

glinton (approved):

PR #468, opened 2022-08-09
chore(cloud-tasks): document scriptID in task find parameters

glinton:
A recent change to cloud added the ability to search for tasks by their referenced script id, this documents that change.

### DIFF
--- a/src/cloud/paths/tasks.yml
+++ b/src/cloud/paths/tasks.yml
@@ -108,6 +108,13 @@ get:
         enum:
           - "basic"
           - "system"
+    - in: query
+      name: scriptID
+      description: |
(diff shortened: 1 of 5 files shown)


src/cloud/paths/tasks.yml
```diff
@@ -108,6 +108,13 @@ get:
         enum:
           - "basic"
           - "system"
+    - in: query
+      name: scriptID
+      description: |
+        Script ID.
+        Only returns tasks using this script.
```
jstirnaman, line 115:
```suggestion
        Only returns tasks that use this script.
```

jstirnaman (approved):

PR #491, opened 2022-08-25
feat: Add endpoint to list clusters

eatondustin1:
This adds a new endpoint spec to list the clusters that are available for the current identity.

### DIFF
--- a/contracts/priv/unity.yml
+++ b/contracts/priv/unity.yml
@@ -221,6 +221,25 @@ paths:
         default:
           description: Unexpected error
           $ref: '#/components/responses/ServerError'
+  /clusters:
+    get:
+      operationId: GetClusters
+      tags:
+        - Clusters
+      summary: Get the list of clusters
+      responses:
+        '200':
+          description: Clusters
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Clusters'
+        '401':
+          description: Unauthorized
+          $ref: '#/components/responses/ServerError'
+        default:
+          description: Unexpected error
+          $ref: '#/components/responses/ServerError'
   /billing:
     get:
       operationId: GetBilling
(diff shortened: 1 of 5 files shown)


contracts/priv/unity.yml
```diff
@@ -1441,6 +1460,42 @@ components:
       required:
         - id
         - name
+    Cluster:
+      properties:
+        isAvailable:
```
randycoulman, line 1465: Do we need separate `isRegionAvailable` and `isProviderAvailable` fields? I don't expect that we'll be adding a new provider any time soon, but the UI has to do different things for unavailable providers than it does for unavailable regions.

I suppose it could infer that, if all regions for a provider are unavailable, then so is the provider, but is that what we want?

contracts/priv/unity.yml
```diff
@@ -1441,6 +1460,42 @@ components:
...
+        priority:
+          description: priority for sorting the cluster
+          nullable: false
+          readOnly: true
+          type: string
+        providerId:
+          description: the id of the cloud provider
+          nullable: false
+          readOnly: true
+          type: string
```
randycoulman, line 1484: Should we have `providerName` in here as well?

randycoulman (approved):
LGTM, but we might need to tweak the list of fields we return.

contracts/priv/unity.yml
eatondustin1: Hmmm, I think you're probably right, but is that something we even want to worry about right now? We could, but we'd probably want to update vault to have both fields too.

contracts/priv/unity.yml
eatondustin1: We could, I just didn't return anything that we weren't currently using, and the way we render the different providers in the UI includes other things that can't come from the API anyway, like the different logos.

contracts/priv/unity.yml
randycoulman: It already does, as does our `Cluster` struct.

contracts/priv/unity.yml
randycoulman: We don't actually use the provider name anywhere in the UI? That surprises me a bit, but if that's the case, then I'm fine with this as-is.

contracts/priv/unity.yml
randycoulman: And, if our existing UI uses the `isProviderAvailable` flag, then I'd like to worry about it right now to preserve existing behavior.

eatondustin1: Well that does makes sense if vault already has it :P I'll update it, thanks!

contracts/priv/unity.yml
eatondustin1: Actually you're completely correct, we do return the provider name, I was only looking at the `extract_region` function. Good catch.

randycoulman (approved):
Changes LGTM!

PR #496, opened 2022-08-29
feat: Add `POST /api/v2private/orgs` to unity

gspletzer:
[Quartz #6659](https://github.com/influxdata/quartz/issues/6659)

In multi-org a user will be able to add an organization to an account in the UI. Due to redirects in the Quartz version of this API, we had to make a separate endpoint for this new feature. This endpoint (`POST /api/v2private`) requires `region`, `provider`, and `orgName` in the request. It will return a `201` with the new organization if successful, otherwise it will return one of the following errors: `400`, `401`, `409`, and `500`.

### DIFF
--- a/contracts/priv/unity.yml
+++ b/contracts/priv/unity.yml
@@ -611,6 +611,38 @@ paths:
         default:
           description: Unexpected error
           $ref: '#/components/responses/ServerError'
+  /orgs:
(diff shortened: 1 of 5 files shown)


abshierjoel (approved):
🙌 👑

Issue #503, opened 2022-08-30
Assign MIT license to spec definitions and repo


jstirnaman:
Add `license` field to `info` sections of specs and to the repo as a whole. Use MIT license. This is approved by InfluxData legal team (https://influxdata.slack.com/archives/C01AMSY5S86/p1660230231759989).

PR #507, opened 2022-08-30
Create LICENSE

sunbryely-influxdata:
Closes https://github.com/influxdata/openapi/issues/503

- Add MIT License to Repo
- Add License Field to Info sections

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2022 InfluxData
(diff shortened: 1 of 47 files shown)


jstirnaman (approved):
Looks good! Unfortunately, we also need to add it to docs-v2/api-docs/openapi/content/*/info.yml files since `info` isn't copied into contracts/ref/*.yml yet.

PR #508, opened 2022-08-31
chore: correct error return for `POST /api/v2private/orgs`

gspletzer:
After additional testing for the new `POST /api/v2private/orgs` was completed, it was noted that Quartz returns `400` when a required parameter is missing, rather than `422`, so the `unity` swagger has been updated to reflect that.

### DIFF
--- a/src/unity/paths/orgs.yml
+++ b/src/unity/paths/orgs.yml
@@ -23,13 +23,17 @@ post:
         Bad request.
         The error may indicate one of the following problems:
         - The request body isn't valid--the request is well-formed, but InfluxDB can't process it due to semantic errors.
-        - You passed a parameter combination that InfluxDB doesn't support
+        - You passed a parameter combination that InfluxDB doesn't support.
       $ref: "../../common/responses/ServerError.yml"
     '401':
       description: Unauthorized. The token passed doesn't have permissions necessary to complete the request.
       $ref: '../../common/responses/ServerError.yml'
     '409':
       description: Conflict. The organization name is already in use.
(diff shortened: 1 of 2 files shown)


src/unity/paths/orgs.yml
```diff
@@ -18,19 +18,19 @@ post:
         application/json:
           schema:
             $ref: '../schemas/OrganizationWithToken.yml'
+    '400':
```
abshierjoel, line 21: So, even if we need the `400`, I think we'll still need the `422` additionally, because `Quartz.provision_organization` can return `{:error, %ValidationError.t()}`, which `FallbackController` turns into a `422`.

abshierjoel (changes requested):
I think we'll need `422` and `400`.

src/unity/paths/orgs.yml
```diff
@@ -18,19 +18,19 @@ post:
...
+      description: |
+        Bad request.
+        The error may indicate one of the following problems:
+        - The request body isn't valid--the request is well-formed, but InfluxDB can't process it due to semantic errors.
+        - You passed a parameter combination that InfluxDB doesn't support
```
sanderson, line 26:
```suggestion
        - You passed a parameter combination that InfluxDB doesn't support.
```

sanderson (commented):
One minor suggestion. Just a missing period.

abshierjoel (approved):

gspletzer:
Quartz code has been modified to eliminate need for `400` on this endpoint; it will only return `422` for invalid request. Closing PR.

PR #533, opened 2022-09-12
fix(user): POST operation for `/users/{userID}/password` doesn't require basic auth

bednar:
Related to https://influxcommunity.slack.com/archives/CHQ5VG6F8/p1662739448171649

The `POST` operation for `/users/{userID}/password` is used by InfluxDB clients (for more info see https://github.com/influxdata/influxdb-client-python/pull/498) and with current `oss.yml` we can't update the user's password.

This PR change the contracts according to current OSS implementation:
- Remove `BasicAuth` from POST operation for `/users/{userID}/password`
- Add `PUT` operation for `/users/{userID}/password`

For more info see:

- https://github.com/influxdata/influxdb/blob/b0a0e734e0c46384499cdc1f08f2673b6a28ff98/tenant/http_server_user.go#L91
- https://github.com/influxdata/influx-cli/blob/3a593e718476d9e671b4a003679722c0b32dc47b/clients/user/user.go#L148

### DIFF
--- a/src/common/paths/users_userID_password.yml
+++ b/src/common/paths/users_userID_password.yml
@@ -20,6 +20,82 @@ post:
 
     - [InfluxDB Cloud - Change your password](https://docs.influxdata.com/influxdb/cloud/account-management/change-password/)
     - [InfluxDB OSS - Change your password](https://docs.influxdata.com/influxdb/latest/users/change-password/)
+  parameters:
+    - $ref: '../parameters/TraceSpan.yml'
+    - in: path
+      name: userID
+      schema:
+        type: string
+      required: true
+      description: The ID of the user to set the password for.
+  requestBody:
+    description: The new password to set for the user.
+    required: true
+    content:
+      application/json:
+        schema:
+          $ref: '../schemas/PasswordResetBody.yml'
+  responses:
+    '204':
+      description: Success. The password is updated.
+    '400':
+      description: |
+        Bad request.
+
+        #### InfluxDB Cloud
+
+        - Doesn't allow you to manage passwords through the API; always responds with this status.
+
+        #### InfluxDB OSS
+
+        - Doesn't understand a value passed in the request.
+      content:
+        application/json:
+          schema:
+            $ref: "../schemas/Error.yml"
+          examples:
+            updatePasswordNotAllowed:
+               summary: Cloud API can't update passwords
+               value: {
+                        "code": "invalid",
+                        "message": "passwords cannot be changed through the InfluxDB Cloud API"
+                      }
+    default:
+      description: Unexpected error
+      $ref: '../responses/ServerError.yml'
+  x-codeSamples:
+    - lang: Shell
+      label: "cURL: use Basic auth to update the user password"
+      source: |
+        curl --request POST \
+           "http://localhost:8086/api/v2/users/USER_ID/password" \
+          --header 'Content-type: application/json' \
+          --user "USERNAME:PASSWORD" \
+          --data-binary @- << EOF
+            {"password": ""}
+        EOF
+put:
+  operationId: PostUsersIDPassword
+  tags:
+    - Security and access endpoints
+    - Users
+  summary: Update a password
+  description: |
+    Updates a user password.
+
+    Use this endpoint to let a user authenticate with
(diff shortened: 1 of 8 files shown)


jstirnaman:
Adding @jeffreyssmith2nd and @lesam just to be sure, but it LGTM - we can improve the docs.

bednar:
The PR is rebased

lesam:
This looks like one for @jeffreyssmith2nd

src/common/paths/users_userID_password.yml
```diff
@@ -18,6 +18,82 @@ post:
...
+    - lang: Shell
+      label: "cURL: use Basic auth to update the user password"
+      source: |
+        curl --request POST \
+           "http://localhost:8086/api/v2/users/USER_ID/password" \
+          --header 'Content-type: application/json' \
+          --user "USERNAME:PASSWORD" \
+          --data-binary @- << EOF
+            {"password": ""}
+        EOF
```
jeffreyssmith2nd, line 76: I don't know that this is correct. Unless I'm missing something, this API only supports token or session auth so this curl will always just bounce back as unauthed.

src/common/paths/users_userID_password.yml
```diff
@@ -18,6 +18,82 @@ post:
...
+put:
+  operationId: PutUsersIDPassword
+  tags:
+    - Security and access endpoints
+    - Users
+  summary: Update a password
+  description: |
+    Updates a user password.
+
+    Use this endpoint to let a user authenticate with
```
jeffreyssmith2nd, line 86: Based on the above, this section is likely incorrect.

src/common/paths/users_userID_password.yml
jstirnaman: > I don't know that this is correct. Unless I'm missing something, this API only supports token or session auth so this curl will always just bounce back as unauthed.

True, doesn't seem to work with Basic Auth.

src/common/paths/users_userID_password.yml
jstirnaman: > Based on the above, this section is likely incorrect.

True, `PUT` with Basic Auth doesn't seem to work.

src/common/paths/users_userID_password.yml
jstirnaman: cURL example with session would be something like:

```sh
  curl -v -c ./cookie-file.tmp --request POST \
   	"$INFLUX_URL/api/v2/signin" \
    --user "${INFLUX_USER_NAME}:${INFLUX_USER_PASSWORD}"


  curl -v --request POST \
    -b ./cookie-file.tmp \
"$INFLUX_URL/api/v2/users/$INFLUX_USER_ID/password" \
    --header 'Content-type: application/json' \
    --data-binary @- << EOF
      {"password": "NEW_USER_PASSWORD"}
EOF
}
```

jstirnaman (changes requested):

src/common/paths/users_userID_password.yml
bednar: Description is updated to:

```
Updates a user password.

    #### InfluxDB Cloud
```

src/common/paths/users_userID_password.yml
bednar: The `POST` example is updated to:

```console
curl --request POST \
           "http://localhost:8086/api/v2/users/USER_ID/password" \
          --header 'Content-type: application/json' \
          --header "Authorization: Token INFLUX_TOKEN" \
          --data-binary @- << EOF
            {"password": "NEW_USER_PASSWORD"}
        EOF
```

and `PUT` example is updated to:

```console
curl -c ./cookie-file.tmp --request POST \
   "http://localhost:8086/api/v2/signin" \
  --user "${INFLUX_USER_NAME}:${INFLUX_USER_PASSWORD}"

curl -b ./cookie-file.tmp --request PUT \
   "http://localhost:8086/api/v2/users/USER_ID/password" \
  --header 'Content-type: application/json' \
  --data-binary @- << EOF
    {"password": "NEW_USER_PASSWORD"}
EOF
```

bednar:
@jstirnaman thanks for review. The description and examples are updated.

jstirnaman (approved):

PR #535, opened 2022-09-13
docs(scripts): add tag content and style fixes for scripts.

jstirnaman:


src/cloud/tags.yml
```diff
@@ -55,20 +55,54 @@ tags:
...
+      Then, InfluxDB executes your script and returns the result.
 
-      Invokable scripts accept parameters.
-      Add parameter references in your script as `params.myparameter`.
-      When you `invoke` your script, you send parameters as key-value pairs in the `params` object.
-      Then, InfluxDB executes your script with the key-value pairs as arguments, and returns the result.
+      To create a script, you can use the [`POST /api/v2/scripts`](#operation/PostScripts) endpoint.
+      After you create a script, extract the script `id` from the response.
+      With the script ID, you can [invoke the script as an API endpoint](#invoke-a-script-as-an-api-endpoint)
+      or [add the script to a task](#add-a-script-to-a-task).
```
kelseiv, line 69: perhaps weave lines 66-69 (which endpoints) into procedural content for brevity in intro

src/cloud/tags.yml
```diff
@@ -55,20 +55,54 @@ tags:
...
+
+      #### Invoke a script as an API endpoint
+
+      Once you create a script, InfluxDB generates an [`/api/v2/scripts/SCRIPT_ID/invoke` endpoint](#operation/PostScriptsIDInvoke) for your
```
kelseiv, line 73: should we have another procedure above to create a script? Either way, think it'd be helpful to be explicit on step for every action they need to take

**Create a script**
1. Use the [`POST /api/v2/scripts`](#operation/PostScripts) endpoint, and do the following:
- Include `params` references in your script--for example, `params.myparameter`
-  (not sure if this is part of creating the script or invoking the script?) Pass `params` with key-value pairs in the response
InfluxDB generates an [`/api/v2/scripts/SCRIPT_ID/invoke` endpoint](#operation/PostScriptsIDInvoke) for your organization.

src/cloud/tags.yml
```diff
@@ -55,20 +55,54 @@ tags:
...
+      organization.
+      To invoke the script, send an HTTP request to the following endpoint:
```
kelseiv, line 75:
```suggestion
      1. Send an HTTP request to 
```

src/cloud/tags.yml
```diff
@@ -55,20 +55,54 @@ tags:
...
+
+      `POST https://INFLUX_CLOUD_URL/api/v2/scripts/SCRIPT_ID/invoke`
+
+      In your request, replace the following:
```
kelseiv, line 79:
```suggestion
      In the request, replace the following:
```

src/cloud/tags.yml
```diff
@@ -55,20 +55,54 @@ tags:
...
+        - **`INFLUX_CLOUD_URL`**: the domain for your InfluxDB Cloud account.
+        - **`SCRIPT_ID`**: the ID of your invokable script.
```
kelseiv, line 81:
```suggestion
        - **`INFLUX_CLOUD_URL`** with the domain for your InfluxDB Cloud account
        - **`SCRIPT_ID`** with the ID of your invokable script
```

src/cloud/tags.yml
```diff
@@ -55,20 +55,54 @@ tags:
...
+
+      To provide parameter values to the script, pass the following in the request body:
+
+        - A `params` object with key-value pairs for parameters referenced in the script
```
kelseiv, line 85:
```suggestion
      Pass the `params` object with key-value pairs in the request body:
```

src/cloud/tags.yml
```diff
@@ -55,20 +55,54 @@ tags:
...
+
+      For more information and examples, see how to [invoke custom scripts as
+      API endpoints]({{% INFLUXDB_DOCS_URL %}}/api-guide/api-invokable-scripts/#create-an-invokable-script).
+
+      #### Add a script to a task
+
+      Once you create a script, you can create or update a task to execute the script.
+
+      To create or update a task, you can use the [`/api/v2/tasks` endpoints](#tag/Tasks).
```
kelseiv, line 94:
```suggestion
      To create or update a task, use the [`/api/v2/tasks` endpoints](#tag/Tasks).
```

kelseiv (approved):

src/cloud/tags.yml
jstirnaman: > 
Following this:
https://developers.google.com/style/placeholders#two-or-more-placeholders

src/cloud/tags.yml
```diff
@@ -55,20 +55,54 @@ tags:
...
+      Pass the following in the request body:
+        - `scriptID`: the ID of the script to execute
+        - `scriptParameters`: an object with key-value pairs for parameters referenced in the script
+
+      For more information and examples, see how to
+      [create a task that references a script]({{% INFLUXDB_DOCS_URL %}}/process-data/manage-tasks/create-task/#create-a-task-that-references-a-script)
```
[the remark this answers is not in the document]
jstirnaman: > to be sure, should we use {{% INFLUXDB_DOCS_URL %}} in most cases we don't know the cloud or oss context?

This file is only processed for Cloud, so it's safe to use the variable here. If it were a common (OSS-Cloud) file then I would use the Cloud-specific URL, but I don't want to if I don't have to.

PR #554, opened 2022-09-22
feat(uiproxyd): add quartz-login-url route

hoorayimhelping:
Adds the `quartz-login-url` route for sending an environment variable to the ui


hoorayimhelping:
This should get merged in after the idpe and k8s-idpe changes

PR #557, opened 2022-09-23
fix(555): finish moving cloud auth into common auth.

### LINKED ISSUE #555: Typescript failure in UI, due to change in oats output.
## Bug:
* occurs with SHA `f84326c9bad6cbdd472fd3baa53da9d60b43aea1`
      * previous SHA is fine.
      * see commit: https://github.com/influxdata/openapi/commit/f84326c9bad6cbdd472fd3baa53da9d60b43aea1
* cloud only errors. oss oats build is fine.

Vid:
 
https://user-images.githubusercontent.com/10232835/192061998-d8a0fec8-d107-4434-97e8-ce6165d5d26e.mov

## Diff in generated files:
* in the UI, we are now missing the typescript definitions for:
   *  Screen Shot 2022-09-23 at 2 58 08 PM

* This is because the generated `contracts/cloud.yml` now has the permissions and resources in lined.
   * was: Screen Shot 2022-09-23 at 2 56 58 PM

   * changed to: Screen Shot 2022-09-23 at 2 57 10 PM
* The oss build does not have this issue, since it does **NOT** have the permissions items in-lined in the `contracts/oss.yml`
   *  Screen Shot 2022-09-23 at 3 13 43 PM

wiedld:
Closes #555 

## Bug:
* failed UI typescript off the oats generated types.
   * In cloud only. 
   * See vid on the issue ticket.
* Failure is due to the generated oats type definitions, having certain nesting certain types.
   * rather than making them explicit type interfaces. (see issue ticket.)
   * But on further digging, the actual error is a bit more.
* In [this commit,](https://github.com/influxdata/openapi/commit/f84326c9bad6cbdd472fd3baa53da9d60b43aea1) the cloud auth was moved to the common auth. But:
  * it had an error for the requestBody schema. Was not pointing at the proper schema.

## Solution:
* fix the 1 line change error.
* Then add the auth request body to the cloud schemas.
* Also, with the new auth schema --> it should also be using the related common schemas for Permissions and Resources.
 

cloud and oss typecheck in UI, now works:
Screen Shot 2022-09-23 at 4 33 15 PM


jstirnaman:
Thanks - sorry for the error!

PR #561, opened 2022-09-29
chore: remove unfunctional property 'oauthID'

glinton:
This removes an apparent unused schema property from both cloud and oss.


jstirnaman:
We're removing instead of deprecating?

glinton:
Whichever you prefer, it was a 5 minute change that felt like we might as well remove it since it's incorrect (rather than do roughly the same amount of work now to mark it as deprecated, just to come along later to remove it :shrug: ).

PR #573, opened 2022-10-13
feat(unity): add query for org creation allowance

randycoulman:
Part of influxdata/quartz#6725

This introduces the concept of an "allowance": is the current user/org/account **allowed** to perform an operation? The response contains the answer to the question as well as additional information that allows a UI to show more useful information.

For now, all we need is a flag indicating whether a plan upgrade would allow an operation that is currently denied, but we can expand this in the future when/if it becomes necessary.

Implementation-wise, an allowance encapsulates any business logic around permitting an operation, including entitlements (aka plan limits) as well as (in the future) roles/permissions.

It also introduces the first allowance query we need for multi-org: can an organization be created?

### DIFF
--- a/contracts/priv/unity.yml
+++ b/contracts/priv/unity.yml
@@ -200,6 +200,31 @@ paths:
         default:
           description: Unexpected error
           $ref: '#/components/responses/ServerError'
+  /allowances/orgs/create:
+    get:
+      operationId: GetAllowancesOrgsCreate
+      tags:
+        - Allowances
+        - Organizations
+      summary: Return whether a new Organization can be created.
+      parameters:
+        - $ref: '#/components/parameters/TraceSpan'
+      responses:
+        '200':
+          description: The Allowance
+          content:
+            application/json:
+              schema:
+                properties:
+                  allowed:
+                    type: boolean
+                    description: is the operation allowed?
+                  upgradesAvailable:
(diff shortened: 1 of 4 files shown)


contracts/priv/unity.yml
```diff
@@ -200,6 +200,31 @@ paths:
...
+      parameters:
+        - $ref: '#/components/parameters/TraceSpan'
+      responses:
+        '200':
+          description: The Allowance
+          content:
+            application/json:
+              schema:
+                properties:
+                  allowed:
```
wdoconnell, line 219: Should the name for this property be generic (`allowed`) or should it be something more specific (like `canCreateOrg`)? Either seems okay to me, but since the response has two booleans, we're kind of providing two pieces of information about what's 'allowed'.

randycoulman: The schema name in the source is `Allowance`, so it's intended to be generic, because this return payload can be reused for any allowance query. Or at least, that's my thinking :-).

The other boolean indicates whether an upgrade might affect the answer to the question of whether something is allowed or not.

wdoconnell: Makes sense to me. 👍

wdoconnell (approved):

hansenreid (approved):

PR #574, opened 2022-10-14
docs: tasks cleanup

jstirnaman:
- Task examples and authorization examples cleanup:
    - replace code samples with request samples where there's nothing
       particularly interesting about the code.
- Move request-specific content into requestBody description.
- Try to apply consistent pattern to params.
- Pattern for describing requestBody. Will add some boilerplate text with
  docs processing.
- Display the task resource schema in the Tasks tag description and add
  a ref that we can link to when referring to properties.
- Refactoring: replace copy-pasted params and properties with  objects.
  Expect more of this, if it meets approval.


src/cloud/paths/dashboards.yml
```diff
@@ -60,19 +60,28 @@ get:
...
       schema:
         type: array
         items:
           type: string
     - in: query
       name: orgID
-      description: The identifier of the organization.
+      description: |
+        An organization ID.
+        Only dashboards that belong to the specified
```
sunbryely-influxdata, line 75:
```suggestion
        Only returns dashboards that belong to the specified
```

src/cloud/paths/dashboards.yml
```diff
@@ -60,19 +60,28 @@ get:
...
+        [organization]({{% INFLUXDB_DOCS_URL %}}/reference/glossary/#organization).
       schema:
         type: string
     - in: query
       name: org
-      description: The name of the organization.
+      description: |
+        An organization name.
+        Only dashboards that belong to the specified
```
sunbryely-influxdata, line 83:
```suggestion
        Only returns dashboards that belong to the specified
```

sunbryely-influxdata (approved):

PR #575, opened 2022-10-18
chore: update clusters contract

subirjolly:
a few properties of `Cluster` entity need to be `boolean` instead of `string`.

### DIFF
--- a/contracts/priv/unity.yml
+++ b/contracts/priv/unity.yml
@@ -1510,22 +1510,22 @@ components:
           description: determines if the cluster is in beta
           nullable: false
           readOnly: true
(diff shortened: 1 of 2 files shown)


eatondustin1 (approved):
Looks good, nice catch!

abshierjoel (approved):
This looks good to me. We're already treating these as `boolean` and `number` types in quartz, so this shouldn't make any conflict 🥳

wdoconnell:
Thanks all!

randycoulman (approved):
This looks right to me. Sorry we got this so wrong :-(.

PR #576, opened 2022-10-18
fix: Use `Organization` schema for create orgs response

abshierjoel:
Fixes the schema being used for the `POST /api/v2private/orgs` endpoint.


wdoconnell (approved):
Looks good to me. Thanks for the quick fix!

PR #577, opened 2022-10-20
chore: Add Provider information to org summaries

abshierjoel:
This adds the following field to the `/api/v2private/accounts/:account_id/orgs` endpoint: `provider`, `region`, and `regionName`.


wdoconnell:
Instead of region, should we call it `regionCode`? That's the property name that gets used when the same data is retrieved from /orgs/:orgId, so might be more consistent.

Example from prod:
provider:"Azure"
regionCode:"westeurope"
regionName:"Amsterdam"

hansenreid (approved):

PR #581, opened 2022-10-24
chore: restore ability to find OSS or Cloud only resource types

DStrand1:
With #557 and #539, the ability to tell if a resource type (see `Resource.yml`) was shared, OSS-only, or Cloud-only. The solution I came up with to avoid reverting all of the work done in those two PRs was to create an unreferenced `Resource.yml` in both `oss/` and `cloud/`, which client libraries (and particularly, `influx-cli`) can reference to determine which permissions are allowed for which platform


jstirnaman:
@DStrand1 Did you try using `allOf` for composition? I think it would make the shared types more obvious. Any reason that won't work in this case? I'll push a branch with an example.

jstirnaman (changes requested):

jstirnaman:
> @DStrand1 Did you try using `allOf` for composition? I think it would make the shared types more obvious. Any reason that won't work in this case? I'll push a branch with an example.

eh, this will work for properties, but not sure it will allow merging `enum` lists.

jstirnaman (approved):

PR #586, opened 2022-10-27
feat(unity): convert upgrade option to enum

randycoulman:
Part of influxdata/quartz#6725

Change the `allowance` payload to return an enum value for what particular upgrade is available rather than just a boolean flag indicating that there is an upgrade available.

This endpoint is not yet in use, so there was no need to change this in an incremental, backwards-compatible way.


gspletzer (approved):
✋🏻 👑 I doth approve

wdoconnell (approved):

PR #591, opened 2022-10-31
refactor: common task properties:

jstirnaman:
- add TaskProperties.yml that defines schemas for properties.
- in _schemas.yml, define NamedSchemas for the properties so they can be referenced elsewhere.
- replace duplicated properties with  to the schemas in TaskProperties.yml
~~- - remove cloud TaskCreateRequest, use the common schema
   change cloud POST /tasks to use oneOf TaskCreateRequest or TaskWithScriptCreateRequest to define the difference in the schema, not just in the description.~~
- cleanup descriptions.

~~Docs example for `oneOf'
image~~

In request examples, we can do this:
image


jstirnaman:
@glinton Please review and let me know if splitting the cloud schema like this will work OK.
Also, if the properties schema works as a pattern.
I don't like having to define each property as a schema individually in _schemas.yml, but that's the only way I could make it work.

src/cloud/paths/tasks.yml
```diff
@@ -212,7 +212,9 @@ post:
     content:
       application/json:
         schema:
-          $ref: "../schemas/TaskCreateRequest.yml"
+          oneOf:
```
jstirnaman, line 215: @glinton Based on the earlier conversation, it sounds like this may not work? It'd be nice to test it and know for sure.

glinton: I wasn't as familiar with the times when it doesn't work, but I see the oats test isn't complaining, maybe when it goes to be used, so I'll leave that bit up to consumers of the spec (this is producing valid swagger).

glinton (approved):

src/cloud/paths/tasks.yml
jstirnaman, line 215: @asalem1 @hoorayimhelping This PR splits the `POST /tasks` request schema into `flux` vs `scriptID` request schemas. Any idea if this `oneOf` will cause a problem for the UI or oats clients? Obviously, it's valid, but I don't want to break implementations that don't support it.

hoorayimhelping: @jstirnaman I don't think `oneOf` will cause problems in and of itself. I see some other examples of it being used in openapi code the UI uses. The biggest risk is the change itself - it looks like this PR adds `TaskWithScriptCreateRequest.yml` in addition to using `TaskCreateRequest.yml`. Changing the description of these parameters and adding a new one might cause issues that aren't obvious right now

jstirnaman: @hoorayimhelping Ok. There isn't anything new here, just explicitly defining the way the API works, but, yeah [the test run](src/tasks/containers/TaskPage.test.tsx) that Gene started for it is failing. Not sure why yet.

hoorayimhelping: @jstirnaman 

```
    src/tasks/containers/TaskPage.test.tsx:259:66 - error TS2339: Property 'flux' does not exist on type 'TaskCreateRequest | TaskWithScriptCreateRequest'.
      Property 'flux' does not exist on type 'TaskWithScriptCreateRequest'.
```

jstirnaman: > @jstirnaman
> 
> ```
>     src/tasks/containers/TaskPage.test.tsx:259:66 - error TS2339: Property 'flux' does not exist on type 'TaskCreateRequest | TaskWithScriptCreateRequest'.
>       Property 'flux' does not exist on type 'TaskWithScriptCreateRequest'.
> ```

I'm assuming it needs a discriminator to know which schema to use? I ran `yarn generate` to generate `src/client/*` locally and I don't even see `TaskWithScriptCreateRequest` in the results, so I 'm confused.
I'll undo this for now and move it to my _someday_ list. It seems beneficial to clients if they could discern which properties are compatible and required. I hate having it human-readable-only fields.

jstirnaman: @sunbryely-influxdata Can you review this for my peace of mind? Because of the failing UI test, I've reverted my change that used `oneOf` for `POST /tasks` schemas in Cloud.

sunbryely-influxdata (approved):

PR #592, opened 2022-10-31
chore(aim): Replace 405s with 403s

abshierjoel:
This updates some Quartz endpoints that return 405 errors to return 403 errors instead.

We were using `Method Not Allowed 405` responses for when a resource is not able to acted upon. I think these errors are better suited at `Forbidden 403`, because the HTTP method is accepted, but the action is not allowed for an internal reason.

### DIFF
--- a/src/unity/paths/orgs_orgId_users_userId.yml
+++ b/src/unity/paths/orgs_orgId_users_userId.yml
@@ -29,6 +29,9 @@ delete:
     '400':
       description: User trying to remove self
       $ref: '../../common/responses/ServerError.yml'
+    '401':
+      description: Unauthorized
+      $ref: '../../common/responses/ServerError.yml'
     '403':
       description: Cannot remove last user
       $ref: '../../common/responses/ServerError.yml'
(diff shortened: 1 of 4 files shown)


src/unity/paths/orgs_orgId_users_userId.yml
```diff
@@ -29,8 +29,8 @@ delete:
     '400':
       description: User trying to remove self
       $ref: '../../common/responses/ServerError.yml'
-    '401':
-      description: Unauthorized
+    '403':
+      description: Cannot remove last user
```
randycoulman, line 33: Isn't it still possible to get a 401 for this endpoint if someone calls it without being authenticated? Should the 403 be additive here, and not a replacement?

randycoulman (changes requested):
Please also get a review from the dumplings team to make sure this isn't going to break anything on their end.

Requested change is to preserve the possible 401 on one endpoint.

src/unity/paths/orgs_orgId_users_userId.yml
abshierjoel: It should still be possible to get a `401`, thanks.

subirjolly (approved):

randycoulman (approved):

PR #601, opened 2022-11-02
chore(unity): remove isBeta flag

randycoulman:
Part of influxdata/quartz#6988

We are eliminating the concept of a beta cluster going forward. This removes the flag from `Cluster` and `OperatorRegion`.

It is not currently used anywhere in the UI codebase (though there is some WIP that might reference it).


subirjolly (approved):
Checked in UI as well. We aren't using it. So, nothing existing should be impacted.

PR #602, opened 2022-11-04
docs: complete merge of OSS and Cloud authorizations paths, describe auths

jstirnaman:
- Merge Cloud /authorizations into common - the only difference is support for ?token= in Cloud.
- Describe the difference.
- Describe required permissions.
- Replace platform-specific /authorizations and /authorizations/ with common.
- Describe patch and delete.
- Cleanup.


jdstrand (approved):
LGTM. AFAICS, the changes are as described which primarily refactor with additional info for a few methods as well as clarifications. Thanks!

PR #603, opened 2022-11-04
chore: use /latest alias for OSS URL substitutions.

jstirnaman:


sanderson (approved):
Approved. Just one question. Will these links persist through to the rendered documentation or are we replacing them with redoc? If they are persisting, this will likely lead to unintended version switches if a user is looking at an older version of the docs and clicks on one of these links. It will switch them to the newest version and they may not realize it.

jstirnaman:
> Approved. Just one question. Will these links persist through to the rendered documentation or are we replacing them with redoc? If they are persisting, this will likely lead to unintended version switches if a user is looking at an older version of the docs and clicks on one of these links. It will switch them to the newest version and they may not realize it.

At first, I was hesitant to use `/latest` for that reason. But now, I think it might be better that we don't have to update the version in openapi. In `docs-v2`, we can rollback (find-and-replace) to the `v[OldVersion]` in the old `docs-v2` contracts as needed when updating. Still too fiddly, though - gotta fix that.

glinton (approved):
I don't feel like I have enough meaningful understanding to say whether this will make a negative impact or not, but the swagger is valid.

PR #617, opened 2022-11-17
chore(aim): Add Org State to unity routes

abshierjoel:
Adds an `OrganizationState` enum and the org state property to the endpoints that return orgs.

### DIFF
--- a/src/unity/schemas/Organization.yml
+++ b/src/unity/schemas/Organization.yml
@@ -25,4 +25,7 @@ properties:
   clusterHost:
     type: string
     description: the url of the current cluster
+  state:
+    $ref: './OrganizationState.yml'
+    description: the current state of the organization
 required: [id, clusterHost, regionCode, regionName]
--- a/src/unity/schemas/Organization.yml
+++ b/src/unity/schemas/Organization.yml
@@ -24,8 +24,8 @@ properties:
     description: provider of the org
   clusterHost:
     type: string
-    description: the url of the current cluster
+    description: URL of the current cluster
   state:
     $ref: './OrganizationState.yml'
-    description: the current state of the organization
+    description: Current state of the organization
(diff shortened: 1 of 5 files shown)


eatondustin1 (approved):
Nice, the enum schema definitely makes more sense, given that we need it for more endpoints than just the operator one.

src/unity/schemas/Organization.yml
```diff
@@ -25,4 +25,7 @@ properties:
   clusterHost:
     type: string
     description: the url of the current cluster
+  state:
+    $ref: './OrganizationState.yml'
+    description: the current state of the organization
```
sanderson, line 30:
```suggestion
    description: URL of the current cluster
  state:
    $ref: './OrganizationState.yml'
    description: Current state of the organization
```

src/unity/schemas/OrganizationState.yml
```diff
@@ -0,0 +1,6 @@
+type: string
+description: state of an organization
```
sanderson, line 2:
```suggestion
description: State of an organization
```

src/unity/schemas/OrganizationSummary.yml
```diff
@@ -21,4 +21,7 @@ properties:
   regionName:
     type: string
     description: the friendly name of the region where the organization is hosted
+  state:
+    $ref: './OrganizationState.yml'
+    description: the current state of the organization
```
sanderson, line 26:
```suggestion
    description: Current state of the organization
```

sanderson (approved):
Approved with minor suggestions

src/unity/schemas/OrganizationSummary.yml
abshierjoel: Thanks for the feedback! I got those changes added!

PR #620, opened 2022-11-29
fix(aim): Add `content-type` to reactivate org endpoint

abshierjoel:
When I added the org reactivate endpoint in #618, I didn't realize Oats would need the content type for an empty body still. This adds that back in so that the `unityRoutes` file gets generated correctly.


eatondustin1 (approved):
🚀

PR #626, opened 2022-12-20
feat: add isDeletable field to account response

eatondustin1:
This adds an isDeletable field to the unity Account schema.


eatondustin1:
> Is there a parallel property on the endpoint that is used in the operator ui?

@wdoconnell yes there is, although it's `deletable` as opposed to `isDeletable`. The operator UI is already using this property.

wdoconnell (approved):

PR #628, opened 2023-01-12
chore(aim): Add IDPE ID to transfer removed user's data to

abshierjoel:
When deleting a user, we will include the option to transfer the user's IDPE data to another user. This adds an optional parameter to the request body to provide a transfer's IDPE ID.

### DIFF
--- a/src/unity/paths/orgs_orgId_users_userId.yml
+++ b/src/unity/paths/orgs_orgId_users_userId.yml
@@ -12,7 +12,7 @@ delete:
           properties:
             transferIdpeId:
               type: string
-              description: the id of the user to transfer IDPE data to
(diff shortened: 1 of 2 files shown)


src/unity/paths/orgs_orgId_users_userId.yml
```diff
@@ -9,6 +9,10 @@ delete:
       application/json; charset=utf-8:
         schema:
           type: object
+          properties:
+            transferIdpeId:
+              type: string
+              description: the id of the user to transfer IDPE data to
```
sanderson, line 15:
```suggestion
              description: User ID to transfer IDPE data to.
```

sanderson (approved):
One minor suggestion.

PR #631, opened 2023-02-02
Revert several changes to fix Influx CLI build

jeffreyssmith2nd:
This PR is meant to serve as a discussion point to see how we might be able to fix the influx cli build. As it stands, the generator tooling used in influx cli cannot handle the current master branch. I believe it has something to do with the new ref style that was added. ```$ref: ./TaskProperties.yml#/status```

I spent about half a day experimenting with different ways to achieve the same idea and couldn't get anything to work, so I wanted to reach out and see if there are any other ideas/suggestions.

With these 2 commits reverted, I can get a working build in Influx CLI


jeffreyssmith2nd:
You can see [here](https://github.com/influxdata/openapi/tree/smith/wip) as far as I got in attempting to refactor, but did not have any luck getting it to build.

sanderson (approved):
LGTM

PR #635, opened 2023-03-08
fix: add missing error code

jeffreyssmith2nd:


wdoconnell:
LGTM, but what's the reason driving the PR - should this error be detectable by the UI somewhere it currently is not?

sanderson (approved):
LGTM

jeffreyssmith2nd:
@wdoconnell OSS is going to start returning not implemented when users attempt to delete by predicate. I don't believe the UI exposes delete functionality, so I don't think there's any immediate impact.

PR #637, opened 2023-05-05
feat: update timerange options for usage

blegesse-git:
Part of #[7238](https://github.com/influxdata/quartz/issues/7238)

This PR updates the `timerange` options for usage information to reflect the latest changes in the system. The endpoint now supports a larger range of `timerange` options, from 1 day (1d or ~24h) to 30 days (30d).


wdoconnell (approved):
Looks good - thanks for adding! Know this is a little tedious to enumerate each option, but I think that's the right approach.

PR #638, opened 2023-05-17
feat: Add cancellable attribute to the operator account object

alejom99:
Adding a new `cancellable` field to the operator account object


blegesse-git (approved):

wdoconnell (approved):
Looks good, and addition of the property shouldn't cause any issues, as I don't think there's any implicit reliance on property order in the operator page. On the UI side I think it'll require a manual mapping in src/operator/constants.tsx

PR #646, opened 2024-02-15
fix: OSS API documentation documents InfluxDB Cloud parameterized queries feature

jstirnaman:
Fixes https://github.com/influxdata/docs-v2/issues/5320 for the OpenAPI contracts. Docs fix pending approval here.

- Effectively removes  property from OSS Query API.
- Re-diverges Query schema, removing  from OSS and keeping it in Cloud.


sanderson (approved):
LGTM 👍🏻

PR #651, opened 2024-05-16
fix: add missing `partial` and schema descriptions for InfluxQL JSON response

### LINKED ISSUE #650: chore: Describe v1 InfluxQL response object schemas
Part of DAR 424

The API dpocumentation for the v1.x compatibility endpoint lists several Array objects under the Series object that are not explained. For example the partial object: the customer is not seeing that key returned in responses but they need to know what (if anything) the field means in order to implement against it.

Reported for InfluxDB Cloud v2.0, but relevant for all versions.

Related:

- v2: https://github.com/influxdata/flux/blob/master/internal/influxql/response.go
- v3: https://github.com/influxdata/influxdb/blob/6f3d6b1b7eaed72a51113ed504b34776e4758000/influxdb3_server/src/http/v1.rs

jstirnaman:
Closes #650, part of https://github.com/influxdata/DAR/issues/424

- Add descriptions for InfluxQL JSON response properties.
-  Add example chunked response
- Fix: `Transfer-Encoding` header to schema (for chunked responses).
- Fix: add missing `partial` property to `results` schema.

### DIFF
--- a/contracts/legacy.yml
+++ b/contracts/legacy.yml
@@ -112,6 +112,15 @@ paths:
                 enum:
                   - gzip
                   - identity
+            Transfer-Encoding:
+              description: |
+                `chunked` if the response is chunked.
+              schema:
+                type: string
+                description: The transfer encoding.
+                default: chunked
+                enum:
+                  - chunked
             Trace-Id:
               description: 'The trace ID, if generated, of the request.'
               schema:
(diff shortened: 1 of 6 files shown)


sanderson (approved):
LGTM

hiltontj (commented):
LGTM - I added https://github.com/influxdata/influxdb/issues/25014 to make sure we are supporting the `partial` parameter in the Edge implementation of the v1 /query API.

hiltontj (approved):
Sorry - meant to ✅ that...

